PR #4, opened 2023-03-17
Clarify time computations are done in POSIX time

### LINKED ISSUE #1: Time math is ambiguously-specified
Leap seconds, UTC, TAI, etc., are all the worst. While we never actually write the timestamps on the wire, it is important that everyone agree, for a given batch number, what times they refer to.

The intent was that you just do math in POSIX time, since that's what every programming language gives you. But if there's ever another leap second, it gets a little tricky. It might be that we have to literally cite POSIX here? Something like: all of these are just measured in seconds, do the math in seconds to get a POSIX timestamp, convert the current time to POSIX time, and then compare in that space?
https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap04.html#tag_04_16

Otherwise you have to answer ridiculous questions like what happens at one of the POSIX timestamps that correspond to multiple times of day.

Not that anything at this granularity matters _at all_ for a PKI.

davidben:
Specifically, that they are done without leap seconds. This ensures the issuance and expiration times of all batches are consistently computed.

I've also moved 'Terminology and Roles' to 'Conventions and Definitions' because it probably should go there.

(This is the same as what would happen if you implemented it naively, but we need to be precise in a spec.)

Fixes #1

### DIFF
--- a/draft-davidben-tls-merkle-tree-certs.md
+++ b/draft-davidben-tls-merkle-tree-certs.md
@@ -42,12 +42,14 @@ normative:
   RFC1123:
 
   X.690:
-       title: "Information technology - ASN.1 encoding Rules: Specification of Basic Encoding Rules (BER), Canonical Encoding Rules (CER) and Distinguished Encoding Rules (DER)"
-       date: February 2021
-       author:
-         org: ITU-T
-       seriesinfo:
-         ISO/IEC 8824-1:2021
+    title: "Information technology - ASN.1 encoding Rules: Specification of Basic Encoding Rules (BER), Canonical Encoding Rules (CER) and Distinguished Encoding Rules (DER)"
+    date: February 2021
+    author:
+      org: ITU-T
+    seriesinfo:
+      ISO/IEC 8824-1:2021
+
+  POSIX: DOI.10.1109/IEEESTD.2018.8277153
 
 informative:
   CHROME-CT:
@@ -225,7 +227,13 @@ when, and only when, they appear in all capitals, as shown here.
 
 This document additionally uses the TLS presentation language defined in {{Section 3 of !RFC8446}}.
 
-# Overview
+## Time {#time}
+
+All time computations in this document are represented by POSIX timestamps, defined in this document to be integers containing a number of seconds since the Epoch, defined in Section 4.16 of {{POSIX}}. That is, the number of seconds after 1970-01-01 00:00:00 UTC, excluding leap seconds. A UTC time is converted to a POSIX timestamp as described in {{POSIX}}.
+
+Durations of time are integers, representing a number of seconds not including leap seconds. They can be added to POSIX timestamps to produce other POSIX timestamps.
+
+The current time is a POSIX timestamp determined by converting the current UTC time to seconds since the Epoch. One POSIX timestamp is said to be before (respectively, after) another POSIX timestamp if it is less than (respectively, greater than) the other value.
 
 ## Terminology and Roles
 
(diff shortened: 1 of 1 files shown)


davidben:
@bob-beck since we've been talking about time a lot lately.

bwesterb (approved):

draft-davidben-tls-merkle-tree-certs.md
```diff
@@ -225,7 +227,13 @@ when, and only when, they appear in all capitals, as shown here.
 
 This document additionally uses the TLS presentation language defined in {{Section 3 of !RFC8446}}.
 
-# Overview
+## Time {#time}
+
+All time computations in this document are represented by POSIX timestamps, defined in this document to be integers containing a number of seconds since the Epoch, defined in Section 4.16 of {{POSIX}}. That is, the number of seconds after 1970-01-01 00:00:00 UTC, excluding leap seconds. A UTC time is converted to a POSIX timestamp as described in {{POSIX}}.
```
bwesterb, line 232: Positive integers I presume.

draft-davidben-tls-merkle-tree-certs.md
```diff
@@ -225,7 +227,13 @@ when, and only when, they appear in all capitals, as shown here.
...
+
+Durations of time are integers, representing a number of seconds not including leap seconds. They can be added to POSIX timestamps to produce other POSIX timestamps.
```
bwesterb, line 234: Do you allow negative integers?

davidben: Hmm, not sure. We don't currently use them anywhere, but it's perfectly defined to add integers in integer space. Dunno, do you have a preference?

draft-davidben-tls-merkle-tree-certs.md
```diff
@@ -225,7 +227,13 @@ when, and only when, they appear in all capitals, as shown here.
...
+
+The current time is a POSIX timestamp determined by converting the current UTC time to seconds since the Epoch. One POSIX timestamp is said to be before (respectively, after) another POSIX timestamp if it is less than (respectively, greater than) the other value.
```
bob-beck, line 236: Should you move this down with the other definitions (issuance time? etc)?

bob-beck (approved):

draft-davidben-tls-merkle-tree-certs.md
davidben: What, you don't want to get a time machine and start a CA in the 1800s? :-)

I suspect this should go the same way as the question below. If we want to be able to say "10 seconds before some time", we should probably allow negative numbers just so the operation is defined everywhere. If not, probably doesn't matter either way.

draft-davidben-tls-merkle-tree-certs.md
bob-beck: remember that time_t is signed.

davidben: Although this spec talks in terms of mathematical integers anyway. But yeah it seems signed is easier here.

draft-davidben-tls-merkle-tree-certs.md
davidben: (Resolving to negatives allowed per other thread. It'll never actually happen but keeps the operations complete.)

draft-davidben-tls-merkle-tree-certs.md
davidben: That's actually the parameters to a CA. The terminology bit is a little awkward... right now, the terminology section just defines enough terms to talk about the overview and the rest is inline. It's... probably not quite the right model. I didn't want to drop a wall of terms without how they go together, but everything inline was differently awkward so... now we have a bit of both. :-/

Thinking I'll leave it as-is for now, since I imagine the organization will need to change more drastically as the document evolves anyway.

PR #28, opened 2023-06-02
Shorten some length prefixes

davidben:
Feedback from Ilari Liusvaara on the list. I picked the lengths mostly arbitrarily. This shaves a few bytes, though it does introduce some length limits over X.509 in TLS today. (TLS uses 2^24-1 for the overall certificate structure. Within a certificate, DER gives variable-length lengths.)


davidben:
Actually, let me close this and do a larger PR with the rest of the comments too. I missed there were a few more.

PR #29, opened 2023-06-02
Shorten some length prefixes

davidben:
Feedback from Ilari Liusvaara on the list. I picked the lengths mostly arbitrarily. This shaves a few bytes, though it does introduce some length limits over X.509 in TLS today. (TLS uses 2^24-1 for the overall certificate structure. Within a certificate, DER gives variable-length lengths.)

### DIFF
--- a/draft-davidben-tls-merkle-tree-certs.md
+++ b/draft-davidben-tls-merkle-tree-certs.md
@@ -344,13 +344,13 @@ enum {
 
 struct {
     ClaimType claim_type;
-    opaque claim_info<0..2^24-1>;
+    opaque claim_info<0..2^16-1>;
 } Claim;
 
 struct {
     SubjectType subject_type;
-    opaque subject_info<0..2^24-1>;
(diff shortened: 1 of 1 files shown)


bwesterb (approved):

davidben:
One thing to ponder: if a ProofType combines X.509-style delegation (so multiple signatures) with one of the really, really large signature schemes, will 64KiB be too tight? I can't imagine actually being happy going that high, but maybe someone wants that? @bwesterb, you probably are better to judge that than me as I've been paying less attention to the sizes available.

bwesterb:
> I can't imagine actually being happy going that high

There are signature schemes with really big public keys, but you wouldn't use those unless you don't have to send the public keys all the time.

At 128 bits security, none of the reasonable schemes will hit the limit. Even SPHINCS+ will not. It has two variants, one optimised for size (~8kB sig, 32B pk) and one for signing speed (~17kB sig, 32B pk).

PR #33, opened 2023-06-05
Properly separate hashes in Merkle tree

### LINKED ISSUE #32: Feedback on hash construction
Feedback from Ilari Liusvaara on the list:

> - The root hash being dependent on issuer and batch number iff there
>  are multiple assertions looks very odd.
>
>  Empty assertion list might be special. But this also happens for
>  one assertion.

It is a little odd that HashNode takes all these extra distinguishers, but HashAssertion doesn't. @bwesterb, you suggested doing something like this to align with something that SPHINCS+ was doing, right? Should we have it on both?

bwesterb:
Fixes #32

Also adds change log.

### DIFF
--- a/draft-davidben-tls-merkle-tree-certs.md
+++ b/draft-davidben-tls-merkle-tree-certs.md
@@ -499,37 +499,57 @@ First, the CA then builds a Merkle Tree from the list as follows:
 Let `n` be the number of input assertions. If `n > 0`, the CA builds a binary tree with l levels numbered `0` to `l-1`, where `l` is `ceil(log_2(n)) + 1`. Each node in the tree contains a hash value. Hashes in the tree are built from the following functions:
 
 ~~~~
-    HashEmpty() = hash(0x00)
+    HashEmpty(level, index) = hash(HashEmptyInput)
     HashNode(left, right, level, index) = hash(HashNodeInput)
-    HashAssertion(assertion) = hash(0x02 || assertion)
+    HashAssertion(assertion, index) = hash(HashAssertionInput)
 ~~~~
 
-`0x00` and `0x02` denote byte strings containing a single byte with value zero and two, respectively. `||` denotes concatenation. `HashNodeInput` is computed by encoding the structure defined below:
+`0x00` and `0x02` denote byte strings containing a single byte with value zero and two, respectively. `||` denotes concatenation.
+
+`HashEmpyInput`, `HashNodeInput` and `HashAssertionInput` are computed by encoding the structures defined below:
 
 ~~~
+struct {
+    uint8 distinguisher = 0;
+    opaque issuer_id<1..32>;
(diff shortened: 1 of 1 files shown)


draft-davidben-tls-merkle-tree-certs.md
```diff
@@ -499,37 +499,57 @@ First, the CA then builds a Merkle Tree from the list as follows:
...
-1. Initialize level 0 with n elements. For `i` between `0` and `n-1`, inclusive,
-   set element `i` to the output of `HashAssertion(assertion[i])`.
+1. Initialize level 0 with n elements. For `j` between `0` and `n-1`, inclusive,
+   set element `j` to the output of `HashAssertion(assertion[j], j)`.
 
 2. For `i` between `1` and `l-1`, inclusive, compute level `i` from level `i-1` as
    follows:
 
-   - If level `i-1` has an odd number of elements, append `HashEmpty()` to the level.
+   - If level `i-1` has an odd number of elements `j`, append `HashEmpty(i, j)` to the level.
```
davidben, line 552: I suppose it doesn't particularly matter in the end, but it is a little annoying that we have to compute these empty hashes. (Does SPHINCS+ do this too?)

Is it worth a section discussing why we need to embed the level and index into everything?

davidben (approved):

draft-davidben-tls-merkle-tree-certs.md
bwesterb: > (Does SPHINCS+ do this too?)
SPHINCS+ only has full binary trees, but if it would have something like HashEmpty, it would contextualise that hash as well.

> Is it worth a section discussing why we need to embed the level and index into everything?
I'll draft something for the security considerations section.

> (I don't fully understand it myself. :-) )
It's to frustrate multi-target attacks.

With the old HashAssertion, any two assertions with the same hash can be used to trick any Merkle Tree CA, included at any time. This is frustrated by the context: an attacker needs to target a specific point in the tree.

Similarly if there are `2^n` nodes across all Merkle Tree CAs, then it only takes `2^(256-n)` tries to find an assertion whose old HashAssertion is one of those hashes.  Probably that's a node in the middle which doesn't make any sense if seen as a whole, but a relying party verifying an authentication path doesn't see the shape of the full tree. `2^n` is roughly the number of non-expired certs.

For HashEmpty I do not see an avenue of attack, but I also haven't convinced myself there's not.

With untruncated 256 bit hashes, all of these attacks would require significant advances in the cryptanalysis of SHA2. We could leave the contextualisation out. It's not very expensive at all, and it makes the scheme much more robust in case the hash turns out to be weaker than expected.

bwesterb: > I'll draft something for the security considerations section.

For tracking: #55

PR #46, opened 2023-06-08
Add storage window 

bwesterb:
It is tempting to add a new state so that an "issued" batch can move
into, say, "oblivion" ("expired" would be confusing.) That would however
distract from the core logic.

We still require CAs to maintain a list of all batches and its
assertions, but don't require them to expose them after the storage
window. I think that's ok.

GH https://github.com/davidben/merkle-tree-certs/issues/2

(Note: forked from #44)

### DIFF
--- a/draft-davidben-tls-merkle-tree-certs.md
+++ b/draft-davidben-tls-merkle-tree-certs.md
@@ -448,6 +448,9 @@ A Merkle Tree certification authority is defined by the following values:
 `validity_window_size`:
 : An integer describing the maximum number of unexpired batches which may exist at a time. This value is determined from `lifetime` and `batch_duration` by `lifetime / batch_duration`.
 
+`storage_window_size`:
+: An integer describing the number of consecutive batches, ending at the latest batch issued, which the CA is guaranteed to serve. `storage_window_size` MUST be at least twice `validity_window_size`.
+
 These values are public and known by the relying party and the CA. They may not be changed for the lifetime of the CA. To change these parameters, the entity operating a CA may deploy a second CA and either operate both during a transition, or stop issuing from the previous CA.
 
 [[TODO: The signing key case is interesting. A CA could actually maintain a single stream of Merkle Trees, but then sign everything with multiple keys to support rotation. The CA -> Subscriber -> RP flow does not depend on the signature, only the CA -> Transparency Service -> RP flow. The document is not currently arranged to capture this, but it probably should be. We probably need to decouple the signing half and the Merkle Tree half slightly. #36 ]]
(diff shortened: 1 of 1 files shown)


draft-davidben-tls-merkle-tree-certs.md
```diff
@@ -434,9 +434,12 @@ A Merkle Tree certification authority is defined by the following values:
 `batch_duration`:
 : A number of seconds which determines how frequently the CA issues certificates. See details below.
 
-`window_size`:
+`validity_window_size`:
 : An integer describing the maximum number of unexpired batches which may exist at a time. This value is determined from `lifetime` and `batch_duration` by `floor(lifetime / batch_duration) + 1`.
 
+`storage_window_size`:
+: An integer describing the number of latest issued batches the CA is guaranteed to serve. `validity_window_size` MUST be at least twice `validity_window_size`.
```
davidben, line 441:
```suggestion
: An integer describing the number batches, ending at the latest batch issued, which the CA is guaranteed to serve. `storage_window_size` MUST be at least twice `validity_window_size`.
```

davidben, line 441: Does a storage window apply to the mirrors too? If not, I don't think it's possible to set up a new mirror for a CA that already exists.

To that end, I think we'd need to update the procedure for mirrors to account for this, as the current procedure assumes the CA is obligated to go back to batch number zero.

bwesterb: 👍

bwesterb: You mean apply to TS? For now we could say a TS is allowed to return 410 as it desires? (And have browsers or another document stipulate stricter requirements for the different use cases of TS?)

davidben: Yeah, I mean that the procedure for the TS, given non-infinite storage windows, is actually kinda tricky. A new TS cannot even run the algorithm in {{single-trusted-service}} because `old_latest_batch` will be zero, and the CA won't serve the old ones.

If a TS is allowed to return 410 whenever it feels like it, that kinda defeats the purpose of a TS I think. The TS is the one that's trusted to _not_ randomly stop serving things. :-)

davidben: Giving them the same storage window seems reasonable. Though whatever we choose will need to be reflected into the update algorithm I mentioned.

bwesterb: Done.

bwesterb:
Rebased.

Issue #5, opened 2023-03-17
Clarify that unparsable signing inputs are still misissuance


davidben:
We currently say:

> If the CA's private key signs an input that can be interpreted as a LabeledWindow structure, the CA is considered to have certified every assertion contained in every value in the `tree_heads` list, with expiry determined by `batch_number`, the position of the tree head in the list, and the CA's input parameters as described in {{parameters}}.

But this leaves open what if the CA's private key signs an input that can't be parsed. Buggy parsers in the RP are bad, but they shouldn't be instant security bugs. To fix that, we should say:

* If the CA's private key signs *any* input beginning with `Merkle Tree Certificate Window\0`, it is in scope for this document. If it cannot be parsed as a LabeledWindow, or if the hashes were invalid, this is misissuance.
* Signing other labels is, a priori, also misissuance, though future protocol extensions MAY define other label prefixes that do have meaning. I.e., if we want to have the CA sign other messages with the same key, we may define a different label for it, but the CA shouldn't be allowed to just randomly sign stuff that seems like fun.

davidben:
It's implied, but we should probably also spell it out explicitly: it doesn't matter if the CA ever served the signature over the HTTP interface. If it generated such a signature, it issued it and is responsible for whatever it signed.

bwesterb:
There is another case: if a signature is served, which itself does not verify, then we should not count it as a misissuance. The signature would not be trusted anyway (as it doesn't verify), and this gives some robustness against some bitflips. #26

davidben:
Oh yeah, there's "I cannot parse the signature" and "you signed a thing that does not parse but has the right context string". I meant the latter. I'll go rename the issue to be clearer.

Agreed that if the signature doesn't verify, it's not misissuance. The system has misbehaved, but I think that can be thought of as a generalized outage. Specifically, it doesn't cause a durable security problem the way having a bad signature in the world does.

PR #47, opened 2023-06-08
Clarify that unparsable signing inputs are still misissuance

bwesterb:
GH #5

### DIFF
--- a/draft-davidben-tls-merkle-tree-certs.md
+++ b/draft-davidben-tls-merkle-tree-certs.md
@@ -1108,7 +1108,7 @@ The second property is achieved by using a collision-resistant hash in the Merkl
 
 ## Cross-protocol attacks {#cross-protocol}
 
-Using the same key material in different, incompatible ways risks cross-protocol attacks when the two uses overlap. To avoid this, {{parameters}} forbids the reuse of Merkle Tree CA private keys in another protocol.  A CA MUST NOT generate signatures with its private key, except as defined in {{signing}}, or an extension of this protocol. Any valid signature of a CA's `public_key` that does not meet these requirements is considered a misissuance.
+Using the same key material in different, incompatible ways risks cross-protocol attacks when the two uses overlap. To avoid this, {{parameters}} forbids the reuse of Merkle Tree CA private keys in another protocol.  A CA MUST NOT generate signatures with its private key, except as defined in {{signing}}, or an extension of this protocol. Any valid signature of a CA's `public_key` that does not meet these requirements indicates misuse of the private key by the CA.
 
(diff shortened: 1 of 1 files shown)


draft-davidben-tls-merkle-tree-certs.md
```diff
@@ -1112,6 +1112,8 @@ Using the same key material in different, incompatible ways risks cross-protocol
 
 To reduce the risk of attacks if this guidance is not followed, the LabeledWindow structure defined in {{signing}} includes a label string, and the CA's `issuer_id`. Extensions of this protocol MAY be defined which reuse the keys, but any that do MUST use a different label string and analyze the security of the two uses concurrently.
 
+Furthermore, any valid signature of a CAs `public_key` that is not of the form as defined in {{signing}}, or an extension of this protocol, MUST be considered a misissuance.
```
davidben, line 1115:
```suggestion
Furthermore, any valid signature of a CA's `public_key` that is not of the form as defined in {{signing}}, or an extension of this protocol, MUST be considered a misissuance.
```

davidben, line 1115: Extremely nitpicky nitpick: "MUST be considered misissuance" feels like kind of a weird directive. (What does it mean if I direct you to "consider" something as something else?) I think something like "CA's MUST NOT do x" and "this is considered a misissuance" reads a bit better. (Then we're just saying that this document considers such behavior a misissuance, as more a definition, than a directive.)

To that end, the previous paragraph says "To avoid this, {{parameters}} forbids the reuse of Merkle Tree CA private keys in another protocol", which I think refers to:

> To prevent cross-protocol attacks, the key used in a Merkle Tree CA MUST be unique to that Merkle Tree CA. It MUST NOT be used in another Merkle Tree CA, or for another protocol, such as X.509 certificates.

{{signing}} also says:

> If the CA's private key signs an input that can be interpreted as a LabeledWindow structure, the CA is considered to have [...]

WDYT if, instead, we appended to the first paragraph something like:

> To avoid this, {{parameters}} forbids the reuse of Merkle Tree CA private keys in another protocol. A CA MUST NOT generate signatures with its private key, except as defined in {{signing}}, or an extension of this protocol. Any valid signature of a CA's `public_key` that does not meet these requirements is considered a misissuance.

(Hmm, is it "misissuance" if my private key gets stolen by an attacker and then the attacker generates a bad certificate, or do we use a different word for that? @devonobrien )

bwesterb: sgtm

davidben: Whoops, @devonobrien and I talked about "misissuance" here and I learned that term is more nuanced than I thought. But I forgot to follow-up here with the conclusion, and now I've forgotten. Devon, do you remember / do you have thoughts on how to say this?

davidben (approved):

draft-davidben-tls-merkle-tree-certs.md
davidben: (Tweaked slightly discussing with Devon.)

Issue #26, opened 2023-05-10
Support multiple logs from a single issuer for incident management


bwesterb:
We should spend a few words on what to do with bit flips.

bwesterb:
I do not think we should cover all this in the draft, but it's a good exercise.

1. **Bitflip in assertion** at CA. Say an assertion is published for `gnogle.com` instead of `google.com`. This should be counted as a misissuance, and detecting these is one of the reasons for CT. I'd say it's out of scope to decide what should happen next: that's a policy matter of the root programs/RPs.
2. **Bitflip when computing authentication path** at CA. The subscriber will get a certificate that simply doesn't validate. The ACME client can check the certificate before installing it. This will be caught early on without lasting impact.
3. **Bitflip when signing a Window** at CA. This one is trickier. On the one hand, when checking the signed window, the TS see that the batch tree heads do not match. They could simply ignore that particular window. However, the situation would be similar to when a CA tries to split the world view. I guess we will need to be tough on this one too. It's not impossible to avoid for the CA: before publishing the signed window, the CA should check whether it verifies to detect any bit flip.
6. **Bitflip when serving assertions** at CA. This is related to the previous. A bit flip might make the CA serve the wrong assertion. I think we can be more lenient here. If the TS fetches the assertions a second time, it'll probably get the correct values and the window signature will validate.
7. **Bitflip when serving a Window** at CA. If the bit flip is in the signature, we can be lenient. It's important that the signed message is explicit: if we would not have included the tree heads in the signed window, then we couldn't be lenient here. If the bit flip is in one of the tree head, we can also be lenient: the signature will not validate and thus we know it's not part of a split world view attack.

davidben:
Post-Photosynthesis triage: Leaving this open to track adding some text about how to recover from policy violations

davidben:
(Still need to write it up, but some of us have been noodling over a "log generation" design to allow CAs to non-destructively cycle to a new log when an existing log goes wrong.)

davidben:
Assigning this to myself. I'll put together a PR for log generations.

davidben:
(Actually @nharper is probably going to do it, but I'm not able to assign it to him on GitHub for some reason.)

nharper:
(I'm not able to assign it to myself, but I'm going to write up log generations aka "the Issuer identifies an entity that can run multiple logs and there's a log number in the upper bits of the serial".)

davidben:
This is done now, forgot to close it.

PR #51, opened 2023-06-12
Add a preliminary section in dealing with bitflips

bwesterb:
See #26


draft-davidben-tls-merkle-tree-certs.md
```diff
@@ -1084,6 +1094,23 @@ However, if the transparency service's interface becomes unavailable, monitors w
...
+1. A signature that does not verify is not considered an misissuance, but rather an availability issue. This allows for bitflips in the signature itself and the signed message.
+2. An (abridged) assertion published via the HTTP interface, is ignored unless it's included in a valid signed window.
+3. A bitflip in a MerkleTreeProofSHA256 can be safely ignored, as any BikeshedCertificate that include it will fail to verify.
+
+There are several places where bitflips cannot be tolerated. For instance, misissuances include (but are not limited to)
+
+1. A bitflip in an assertion, which is included (with bitflip) in a valid signed window.
+2. A valid signed window, for which the CA cannot *eventually* produce assertions that lead to the included tree heads.
+
+CAs should take care to catch these bitflips. For instance, a CA should double-check the computation of tree heads before signing them.
```
devonobrien, line 1112: Given CT's ambiguity here left us in a bad spot wrt bit flips, could we refer to a strong practice or just prescribe specific steps here? e.g. Sign --> verify --> Publish or Sign1 --> Sign2 --> Compare --> Publish. Catching these before the signatures get into the wild really does reduce it to a pure availability problem.

bwesterb: I'm happy to add a list of suggested countermeasures. However, I don't want to seem exhaustive.

draft-davidben-tls-merkle-tree-certs.md
```diff
@@ -1100,6 +1110,23 @@ However, if the transparency service's interface becomes unavailable, monitors w
...
+Where security allows it, the present system has been designed to be lenient with bitflips. For example:
+
+1. A signature that does not verify is not considered an misissuance, but rather an availability issue. This allows for bitflips in the signature itself and the signed message.
+2. An (abridged) assertion published via the HTTP interface, is ignored unless it's included in a valid signed window.
+3. A bitflip in a MerkleTreeProofSHA256 can be safely ignored, as any BikeshedCertificate that include it will fail to verify.
+
+There are several places where bitflips cannot be tolerated. For instance, misissuances include (but are not limited to)
+
+1. A bitflip in an assertion, which is included (with bitflip) in a valid signed window.
+2. A valid signed window, for which the CA cannot *eventually* produce assertions that lead to the included tree heads.
```
davidben, line 1126: Aren't the two cited bitflips really just analogs to these two failures? I.e., at least as currently written, we aren't any more bitflip-lenient than CT was. There is potentially some room for bitflip leniency purely on grounds that the bust trees will eventually age out, and you could imagine the TSes just refusing to forward the bust tree heads, but that feels like something different and may require some more normative text.

bwesterb: > There is potentially some room for bitflip leniency purely on grounds that the bust trees will eventually age out, and you could imagine the TSes just refusing to forward the bust tree heads

It's scary to be lenient like this, because it threatens transparency. We don't want a CA to just show a particular batch to only one tiny TS.

> Aren't the two cited bitflips really just analogs to these two failures?

Yes and no.

The yeti2022 bit flip indeed is an example of 2, and we can't be lenient with it. The bit flip would've been caught, if the CT log would've recomputed the tree head just to be sure, as we suggest in the text.

The nessie2024 bit flip is a bit funny. It's not quite 1. The problem was that the SubjectPublicKeyInfo in the tbs_certificate didn't match the one in the pre_certificate because of a bit flip. With MTC there is no duplication: we'd just have a bit flip in the subject_info_hash. It's the subscriber's job to notice the error. If they see that hash only differs by one bit from the intended hash, then that's technically a misissuance, but one without any potential for a security breach, and we can be lenient.

davidben: > It's scary to be lenient like this, because it threatens transparency. We don't want a CA to just show a particular batch to only one tiny TS.

Even the tiny TS will detect this because the input assertions will not match the signature at the top. Your TS, large or small, should be unwilling to pass tree heads along if they weren't able to build the tree.

But, yeah, it would be a whole lot of engineering and policy work when the alternative is pretty straightforward:
* The update process will naturally fail to work and so the CA -> TS -> RP flow will freeze. No new certs will work but old ones are still fine
* Recovery can simply be winding down the old CA and making a new one

> It's the subscriber's job to notice the error. If they see that hash only differs by one bit from the intended hash, then that's technically a misissuance, but one without any potential for a security breach, and we can be lenient.

I dunno how much we want to play the bit-flip-counting game here (also the bitflip could have happened in the middle of computing the hash, in which case it'll be a bit more of a mess). But my actual point is that nothing in this PR is special to MTCs.

* In both systems, if you thought you signed something but didn't, you didn't do anything. Although, more on this below.
* In both systems, if you serve some garbage instead of cryptographically coherent log state, it's an availability issue in your serving frontend
* In both systems, if you issue and log a certificate just fine, but send some other byte string to the subscriber, that's between you and the subscriber. Try again at delivering your byte string.
* In both systems, if you issue a certificate asserting a false thing, that's unambiguously misissuance. It's up to root programs to make a policy decision what to do here.
* In both systems, if you sign the wrong or a garbage tree head, that's unambiguously a failure in the log. MTCs can contain the impacts a bit better just because each batch is independent, but from the above discussion it sounds like we're not too excited about leaning on that anyway
* In both systems, double-checking computation can help if your computing environment is prone to bit flips

So I think we aren't _actually_ "designed to be lenient with bitflips". Which I also think is fine. It is hard to quantify the impact of a bitflip. Leaving it to a case-by-case policy decision seems the right path anyway. For example, the first claim ("A signature that does not verify...") is very very dangerous. If your bitflip caused you branch into the DoNothing() function instead of the SignSomething() function, that's fine. But more likely your bitflip messed up some internal computation in your signing algorithm, in which case we'd need to dig into the literature on fault attacks for your algorithm. You may well have burned your private key.

So I'm inclined to say we just don't say anything in the draft and leave it be.

bwesterb: > Even the tiny TS will detect this because the input assertions will not match the signature at the top. Your TS, large or small, should be unwilling to pass tree heads along if they weren't able to build the tree.

💯 

> But, yeah, it would be a whole lot of engineering (punching holes in the update flow) and policy work when the alternative is pretty straightforward:
> 
> * The update process will naturally fail to work and so the CA -> TS -> RP flow will freeze. No new certs will work but old ones are still fine

This is sensible.

> * In both systems, double-checking computation can help if your computing environment is prone to bit flips

The nessie2024 bit flip would've been prevented if the computation was double checked. Whether some in an RFC will ensure all implementations will be thoughtful of bitflips is a different question.

> So I think we aren't _actually_ "designed to be lenient with bitflips".

I'm ok leaving that out. It's more important that we actually thought through where bitflips can happen and how we should deal with them.

davidben:
PR itself no longer applies with Photosynthesis, but we might perhaps include a discussion about how to recover from policy violations.

PR #52, opened 2023-06-14
Minor editorial changes

thomwiggers:
* Separate the list of allowed/not allowed values for `dnsname` by regular text so that `xn--iv8h.example` and `EXAMPLE.com` don't blur together
* Format the `*.` string
* Remove an unnecessary comma


davidben (approved):
Thanks!

Issue #55, opened 2023-06-19
Expand upon security requirements of hash in security consideration section


bwesterb:
Cover:

- Hash contextualisation, cf #32 
- Where collision resistance is really used #45

davidben:
Post-Photosynthesis triage: security considerations mentions depending on collision-resistance. Is there more stuff we should do here with a CT-like tree?

davidben:
Think we can close this? The public key is now clarified as a second-preimage situation. The other stuff cites collision-resistance, but that's true for now. We have the hooks in place to define a randomizer if we want, and then that text will want to be updated with it.

Issue #45, opened 2023-06-06
Is it worthwhile to avoid requiring collision resistance from our hash?


bwesterb:

davidben:
With a CA-provided randomizer, we'd still rely on second preimage resistance for transparency, right? (I.e. we couldn't halve the hash size.) Though I suppose we wouldn't rely on it for authenticity.

bwesterb:
Yes, you're right: even with the CA-provided randomiser, we would still rely on *collision resistance* to prevent the CA from providing two world views. Indeed, if we'd truncate hashes to 16 bytes, then with 2^64 work, a CA could find two different batches with the same root hash.

I guess we couldn't get rid of that without significant modifications (see #48). Still I think it might be worthwhile to reduce reliance of collision resistance.

davidben:
Post-Photosynthesis triage: eh, I guess we've decided to just import a CT/tlog-style tree structure. Probably can close this.

bwesterb:
Let's [reopen](https://mailarchive.ietf.org/arch/msg/plants/FN8L2-C2AjN7QxpX6RC7b3tnxi8/) this.

PR #57, opened 2023-06-20
Add batch randomizer and elaborate on hash requirements

bwesterb:
Cf #45 #55

This does add 16 bytes to each proof. Is it worth it? Perhaps something to discuss in the working group.


bwesterb:
Rebased.

davidben:
Closing this with Photosynthesis since we're now just using tlog for the tree construction.

Issue #12, opened 2023-03-22
Flesh out HTTP interface


davidben:
Filing this to capture the TODO left in the doc

bwesterb:
> [[TODO: This section hasn't been written yet. For now, this is just an informal sketch. The real text will need to define request/response formats more precisely, with MIME types, etc. ]]

> [[TODO: If we have the window endpoint, do we still need to separate "info" and "assertions"?]]

davidben:
Moot with Photosynthesis. tlog gives us an HTTP API.

PR #59, opened 2023-06-22
Start fleshing out HTTP interface

bwesterb:
See #12.

I tried to keep it as simple as possible.

Needs to be updated after we add the storage window. #46 
Signatures are defined in #50 
Needs to be updated after merging #46, #57 and #49

### DIFF
--- a/draft-davidben-tls-merkle-tree-certs.md
+++ b/draft-davidben-tls-merkle-tree-certs.md
@@ -448,6 +448,9 @@ A Merkle Tree certification authority is defined by the following values:
 `validity_window_size`:
 : An integer describing the maximum number of unexpired batches which may exist at a time. This value is determined from `lifetime` and `batch_duration` by `lifetime / batch_duration`.
 
+`http_server`:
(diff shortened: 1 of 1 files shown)


bwesterb:
Rebased.

draft-davidben-tls-merkle-tree-certs.md
```diff
@@ -448,6 +448,9 @@ A Merkle Tree certification authority is defined by the following values:
 `validity_window_size`:
 : An integer describing the maximum number of unexpired batches which may exist at a time. This value is determined from `lifetime` and `batch_duration` by `lifetime / batch_duration`.
 
+`http_server`:
+: A servername together with optional port and path prefix, under which the CA publishes, as described in {{publishing}}. For example: `example.com/ca`.
```
lukevalenta, line 452: If we go with the `tlog` approach (#103), we probably want to make this the same as the submission prefix/origin line from https://github.com/C2SP/C2SP/blob/main/static-ct-api.md#checkpoints.

davidben:
Closing this with Photosynthesis since we're now just using tlog for the HTTP interface.

Issue #31, opened 2023-06-02
Canonicalize claim structures further and think about multiplicity


davidben:
Feedback from Ilari Liusvaara on the list:

>  I don't see rule for how claims are sorted within each type,
>  only how different types are sorted.

and:

> - If each claim was in its own Claim, then one could maybe even
>  shorten [the length prefix] to 8 bits. Similarly, one could merge ipv4/ipv6 and
>  dns/dns_wildcard.
>
>  This could also simplify sorting: Sort by type, lexicographic
>  sort by claim contents.

Moving the multiplicity up a layer avoids every ClaimType defining it itself, and reuses some length prefixes given how the TLS presentation language works, though it means a non-multiplicative Claim would need extra logic to reject duplicates. Is that a problem?

One way or another, I am a fan of canonicalizing the order more strongly, just for the sake of having only one way to encode it. Though how we canonicalize may be interesting... if DNS names are pre-sorted in hierarchical order, evaluating name constraints efficiently becomes a lot easier.

davidben:
Post-Photosynthesis triage: X.509 is itself completely deficient here but we've tabled BikeshedCertificate for now.

PR #62, opened 2023-07-04
Specify canonical representation for assertions

bwesterb:
GH #31

### DIFF
--- a/draft-davidben-tls-merkle-tree-certs.md
+++ b/draft-davidben-tls-merkle-tree-certs.md
@@ -391,6 +391,35 @@ DNSName values use the "preferred name syntax" as specified by {{Section 3.5 of 
 
 Names in a `dns` claim represent the exact DNS name specified. Names in a `dns_wildcard` claim represent wildcard DNS names and are processed as if prepended with the string "`*.`" and then following the steps in {{Section 6.3 of !I-D.draft-ietf-uta-rfc6125bis}}.
 
+### Canonical order {#dnsname-order}
+
+The values in `dns_names` MUST be sorted lexicographically by label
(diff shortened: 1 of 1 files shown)


draft-davidben-tls-merkle-tree-certs.md
```diff
@@ -391,6 +391,35 @@ DNSName values use the "preferred name syntax" as specified by {{Section 3.5 of
 
 Names in a `dns` claim represent the exact DNS name specified. Names in a `dns_wildcard` claim represent wildcard DNS names and are processed as if prepended with the string "`*.`" and then following the steps in {{Section 6.3 of !I-D.draft-ietf-uta-rfc6125bis}}.
 
+### Canonical order {#dnsname-order}
+
+The values in `dns_names` MUST be sorted lexicographically by label
+from right-to-left; that is:
```
bwesterb, line 397: Also equivalent to:

```python
def mtcSort(l):
    return ['.'.join(reversed(x))
            for x in sorted([tuple(reversed(x.split('.'))) for x in l])]
```

draft-davidben-tls-merkle-tree-certs.md
```diff
@@ -391,6 +391,35 @@ DNSName values use the "preferred name syntax" as specified by {{Section 3.5 of
...
+
+The following is a list of properly sorted names in this order:
+
+~~~
+example.com
+b.example.com
+b.b.example.com
+ba.example.com
+z.example.com
+~~~
```
bwesterb, line 421: This list is chosen such that normal sorting or sorting the reversed words will not work. Also it catches all cases in the definition.

Issue #66, opened 2023-09-07
Avoid floating points


bwesterb:
It's tempting to compute `ceil(log_2(n))+1` using floating points. Even though I doubt we'll see batches of around 2^53 certs, it's better to stick to integers.

PR #68, opened 2023-09-07
Avoid the temptation of floating points

bwesterb:
GH #66

### DIFF
--- a/draft-davidben-tls-merkle-tree-certs.md
+++ b/draft-davidben-tls-merkle-tree-certs.md
@@ -496,7 +496,7 @@ This section describes how to certify a given list of assertions at a given batc
 
 First, the CA then builds a Merkle Tree from the list as follows:
 
-Let `n` be the number of input assertions. If `n > 0`, the CA builds a binary tree with l levels numbered `0` to `l-1`, where `l` is `ceil(log_2(n)) + 1`. Each node in the tree contains a hash value. Hashes in the tree are built from the following functions:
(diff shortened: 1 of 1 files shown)


draft-davidben-tls-merkle-tree-certs.md
```diff
@@ -496,7 +496,7 @@ This section describes how to certify a given list of assertions at a given batc
 
 First, the CA then builds a Merkle Tree from the list as follows:
 
-Let `n` be the number of input assertions. If `n > 0`, the CA builds a binary tree with l levels numbered `0` to `l-1`, where `l` is `ceil(log_2(n)) + 1`. Each node in the tree contains a hash value. Hashes in the tree are built from the following functions:
+Let `n` be the number of input assertions. If `n > 0`, the CA builds a binary tree with l levels numbered `0` to `l-1`, where `l` is the least positive integer with `n <= 2^(l-1)`. Each node in the tree contains a hash value. Hashes in the tree are built from the following functions:
```
davidben, line 499: Very nitpicky nitpick: this reads a little clearer to me, but I don't know if this is actually the right way to phrase this, or the particular flavor of math jargon that I learned.

```suggestion
Let `n` be the number of input assertions. If `n > 0`, the CA builds a binary tree with l levels numbered `0` to `l-1`, where `l` is the smallest positive integer such that `n <= 2^(l-1)`. Each node in the tree contains a hash value. Hashes in the tree are built from the following functions:
```

davidben (approved):

Issue #72, opened 2023-09-08
No need to be clever with the pad


bwesterb:
In the current draft we carefully add some padding in the hash, so that we can reuse the computation of the first block among many hashes. From @mia1024's PoC in Python it's clear that hashing is far from the bottleneck, but that the pad is annoying to deal with. Let's remove the pad.

PR #73, opened 2023-09-08
Simplify hashing by not padding to hash block size

bwesterb:
Hashing is very fast already: we do not need the speed up, and the padding made it a bit harder to implement.

GH #72

### DIFF
--- a/draft-davidben-tls-merkle-tree-certs.md
+++ b/draft-davidben-tls-merkle-tree-certs.md
@@ -522,7 +522,6 @@ struct {
     uint8 distinguisher = 0;
     opaque issuer_id<1..32>;
     uint32 batch_number;
-    opaque pad[N];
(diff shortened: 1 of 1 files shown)


davidben:
> Hashing is very fast already

In other words, never use MTCs with SHA-3? 😉

davidben (approved):

davidben:
Probably the right padding construction would be

```
struct {
    opaque issuer_id<1..32>;
    uint32 batch_number;
    opaque pad[N];
    uint8 distinguisher = ...;
    ...
```

That way you can feed the first block into your hash context once and then clone the hash context for all the subsequent bits. Previously we needed to freeze three different ones.

bwesterb:
> In other words, never use MTCs with SHA-3? 😉

shush!

Issue #76, opened 2023-10-12
AbridgedAssertion does not contain SignatureScheme


bwesterb:
Is this an issue?

We could instead define AbridgedAssertion as

```
struct {
    SubjectType subject_type;
    opaque abridged_subject_info<0..2^16-1>;
    Claim claims<0..2^16-1>;
} AbridgedAssertion;

struct {
    SignatureScheme signature;
    opaque public_key_hash[hash.length];
} AbridgedTLSSubjectInfo
```

where obviously `abridged_subject_info` has to be `AbridgedTLSSubjectInfo` in case of subject_type=`tls`.

davidben:
I guess the tradeoff is we avoid redacting things we don't need, but now the verification process needs to know about TLS.

I don't think redacting it is a *problem* per se, though I also don't know how much weight to put on this separation. My intuition is the separation is useful but dunno how it'll shake out.

Issue #38, opened 2023-06-05
Should TLSSubjectInfo have an extension list?


bwesterb:

bwesterb:
If the extension is critical, then it is undeployable without some form of negotiation. If there is negotiation, then instead we can simply negotiate a `TLSSubjectInfo2` via `bikeshed_tls2`, cf #7.

So, which non-critical extensions could we imagine requiring in the future?

davidben:
The one (very, very speculative) use case I keep tossing around my head is a downgrade protection scheme.

Imagine if some site needs to mint both `TLSSubjectInfo` and `TLSSubjectInfo2`. Let's say we designed this back in time and `TLSSubjectInfo` is like RSA key exchange or SSL 2.0 or something horrid like that. As a server, you don't want modern clients to be vulnerable to those algorithms. But the old clients aren't gone yet, so you still need to have `TLSSubjectInfo` around. Conversely, the old servers aren't all upgraded, so the new clients still need to support the old protocols.

Or even within `TLSSubjectInfo`, perhaps we're in a PQ transition state where you have both an RSA TLS key and a Dilithium TLS key, and clients support both.

We'd like for the modern clients to reject the modern server's weak credential, because they should only ever be presented with the strong credential. One way to solve this is a non-critical extension (or, equivalently, a critical extension we define today) that means "distrust this if you understand TLSSubjectInfo2" or "distrust this if you understand Dilithium". The old clients would ignore this, while the new clients would recognize it and follow the instructions.

(All that said, this could just be a Claim rather than a field inside TLSSubjectInfo. It's just kinda weird because the CA isn't really validating anything about this Claim. Their only responsibility is to faithfully put in what the subscriber wanted.)

bwesterb:
> The one (very, very speculative) use case I keep tossing around my head is a downgrade protection scheme.

That's a very good idea. I think we should include that by default. #54

PR #77, opened 2023-10-13
Split subject_info into ~ and abridged_subject_info

bwesterb:
GH #76

### DIFF
--- a/draft-davidben-tls-merkle-tree-certs.md
+++ b/draft-davidben-tls-merkle-tree-certs.md
@@ -538,7 +538,7 @@ struct {
 
 struct {
     SubjectType subject_type;
-    opaque subject_info_hash[hash.length];
+    opaque abridged_subject_info<0..2^16-1>;
     Claim claims<0..2^16-1>;
 } AbridgedAssertion;
 
@@ -551,7 +551,7 @@ struct {
 } HashAssertionInput;
 ~~~
 
-`issuer_id` and `batch_number` are set to the CA's `issuer_id` and the current batch number. `HashAssertionInput.abridged_assertion.subject_info_hash` is set to `hash(assertion.subject_info)` from the function input `assertion`, and the remaining fields of `HashAssertionInput.abridged_assertion` are taken unmodified from `assertion`. The remaining fields, such as `index`, are set to inputs of the function.
+`issuer_id` and `batch_number` are set to the CA's `issuer_id` and the current batch number. `abridged_assertion.abridged_subject_info` is set to `abridge_subject_info(assertion.subject_type, assertion.subject_info)` defined in {{tls-subject-info}} from the function input `assertion`, and the remaining fields of `abridged_assertion` are taken unmodified from `assertion`. The remaining fields, such as `index`, are set to inputs of the function.
 
 Tree levels are computed iteratively as follows:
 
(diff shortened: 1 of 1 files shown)


davidben:
Sorry, I realized https://github.com/davidben/merkle-tree-certs/issues/76#issuecomment-1759724011 was probably ambiguous. I meant I think the it's useful to have a logical separation between the TLS-specific TLSSubjectInfo format and the (so far) SubjectType-agnostic MTC format. E.g. it means you can define a new SubjectType without having to update every MTC implementation.

Though it does mean the log doesn't tell you everyone's key types. Dunno how much to weigh that. WDYT?

bwesterb:
> E.g. it means you can define a new SubjectType without having to update every MTC implementation.

I don't quite follow. The difference is that with this change, you need an up-to-date implementation to abridge an assertion.

TSes and monitors don't abridge (they only deal with already abridged assertions); CAs shouldn't deal with subjects they don't support; and subscribers don't care about subjects they don't support. That leaves the relying party: it can't verify the proof in a BikeshedCertificate without being updated. Does that matter though: presently it can, but what use is that proof if part of the statement proven (subject)  is not understood anyway.

davidben:
Ah that's true. The TS doesn't actually care. Still, imagine if you're using a BikeshedCertificate verifier from the OS or so. Or even if not from the OS, some separate library. A SubjectType-agnostic abridging process would mean you can freely introduce new subject types (i.e. new protocols) without any fuss.

As an analogy, we don't need to patch our X.509 verifiers to introduce new EKUs. (Although EKUs are kind of a mess for other reasons, so new protocols tend not to make new ones. That's not great for cross-protocol purposes. Would be nice to avoid those reuse pressures.)

bwesterb:
> Still, imagine if you're using a BikeshedCertificate verifier from the OS or so.

Fair point. The user can pass an AbridgedMerkleTreeCertificate defined in the obvious way though.

Note that we do have to change the OS certificate API to support MTC, as the relying party needs to fetch the latest available batches for negotiation. Thus we have some extra flexibility there if you don't like abridging the cert.

A different path is to split SubjectInfo in two in a different place. Have a HiddenTLSSubjectInfo, and a TransparentTLSSubjectInfo. Then we define a (unabridged) SubjectInfo as a pair of them, and an AbridgedSubjectInfo as the hash of the former, and the full info of the latter. Conceptually this is elegant, but it feels like unnecessary complexity and is messy to present.

davidben:
Right, so I guess the question is how much of a big deal is it that the whole SubjectInfo gets hashed right now. If it's not that big of a deal (we just lose the key type), just hashing the whole thing is most straightforward.

This is roughly analogous to whether you hash the whole SPKI in X.509 or just the BIT STRING inside it.

draft-davidben-tls-merkle-tree-certs.md
```diff
@@ -538,7 +538,7 @@ struct {
 
 struct {
     SubjectType subject_type;
-    opaque subject_info_hash[hash.length];
+    opaque abridged_subject_info<0..2^16-1>;
```
pohlm01, line 541: Is there a reason this is a (length prefixed) array? If I'm not mistaken, the `abridged_subject_info` has a fixed length implied by the CA parameters hash length, so don't see a need for the length prefix. I also don't see a reason to put multiple elements.

pohlm01:
@bwesterb and @davidben did you have the chance to develop an opinion on whether this PR will likely be merged? [bwesterb/mtc](https://github.com/bwesterb/mtc) currently implements this PR, and I wonder if I should follow this for [my implementation](https://github.com/pohlm01/mct-verifier).

draft-davidben-tls-merkle-tree-certs.md
bwesterb: For TLS it's fixed length at the moment. For other subjects types it might not be fixed length; and for TLS it might even change, see #38.

pohlm01: Ah, thanks for clarification!

davidben:
Closing this with Photosynthesis.

PR #87, opened 2024-07-26
Rebase the draft on top of draft-beck-tls-trust-anchor-ids

### LINKED ISSUE #80: Rebase negotiation scheme on top of trust anchor IDs
We can probably also trim the now redundant text around multi-certificate deployment models.

davidben:
The main changes:

* ProofType is gone and now implicit from the trust anchor

* TrustAnchor is gone and replaced with TrustAnchorIdentifier from draft-beck-tls-trust-anchor-ids.

* The negotiation mechanism is removed and instead references draft-beck-tls-trust-anchor-ids, adding some guidance on how to use it.

* We did lose the aliases mechanism and instead take a harder dependency on the DNS mechanism in draft-beck-tls-trust-anchor-ids. I've filed https://github.com/davidben/tls-trust-expressions/issues/62 to track restoring that.

* General discussion on agility is removed as draft-beck-tls-trust-anchor-ids covers that.

* Just to get it out of the way, I moved the discussion on client certificate types to an appendix.

Fixes #80

### DIFF
--- a/draft-davidben-tls-merkle-tree-certs.md
+++ b/draft-davidben-tls-merkle-tree-certs.md
@@ -301,7 +301,7 @@ The process of issuing and using a certificate is as follows:
 
 6. In an application protocol such as TLS, the relying party communicates its currently saved validity window to the subscriber.
 
-7. If the relying party’s validity window contains the subscriber’s certificate, the subscriber negotiates this protocol and sends the Merkle Tree certificate. See {{trust-anchor-negotiation}} for details. If there is no match, the subscriber proceeds as if this protocol were not in use (e.g., by sending a traditional X.509 certificate chain).
+7. If the relying party’s validity window contains the subscriber’s certificate, the subscriber negotiates this protocol and sends the Merkle Tree certificate. See {{certificate-negotiation}} for details. If there is no match, the subscriber proceeds as if this protocol were not in use (e.g., by sending a traditional X.509 certificate chain).
 
 {{fig-deployment}} below shows this process.
 
@@ -431,7 +431,7 @@ A Merkle Tree certification authority is defined by the following values:
 : A cryptographic hash function. In this document, the hash function is always SHA-256 {{SHS}}, but others may be defined.
 
 `issuer_id`:
-: An opaque byte string that identifies the CA. This value should be short and is limited to at most 32 bytes.
+: A trust anchor identifier (see {{Section 3 of !I-D.beck-tls-trust-anchor-ids}}) that identifies the CA. See {{identifying}} for details.
 
 `public_key`:
 : The public half of a signing keypair. The corresponding private key, `private_key`, is known only to the CA.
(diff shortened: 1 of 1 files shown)


draft-davidben-tls-merkle-tree-certs.md
```diff
@@ -716,60 +712,36 @@ This section describes how subscribers present and relying parties verify Merkle
...
-## Certificate Negotiation {#trust-anchor-negotiation}
-
-Merkle Tree certificates can only be presented to up-to-date relying parties, so this document describes a mechanism for subscribers to select certificates. This section describes the general negotiation mechanism. {{tls-trust-anchors-extension}} describes it as used in TLS.
+4. If `remaining` is non-zero, abort this procedure with an error.
 
-Subscribers maintain a certificate set of available BikeshedCertificates. The TrustAnchor value in each BikeshedCertificate is known as the primary TrustAnchor. Each BikeshedCertificate is also associated with the following values:
+5. If `hash` is not equal to the batch's tree head in the relying party's saved validity window (see {{relying-parties}}), abort this procedure with a `bad_certificate` error.
 
-* A set of additional TrustAnchor values which also match this certificate
+6. Optionally, perform any additional application-specific checks on the assertion and issuer. For example, an HTTPS client might constrain an issuer to a particular DNS subtree.
```
davidben, line 742: Self-review: this and step 7 should be part of the generic process, not the issuer-specific process.

bob-beck (approved):

draft-davidben-tls-merkle-tree-certs.md
```diff
@@ -626,18 +634,13 @@ A CA MUST NOT generate signatures over inputs that are parseable as LabeledValid
...
-enum { merkle_tree_sha256(0), (2^16-1) } ProofType;
-
 struct {
-    ProofType proof_type;
-    opaque trust_anchor_data<0..2^8-1>;
-} TrustAnchor;
-
-struct {
-    TrustAnchor trust_anchor;
+    TrustAnchorIdentifier trust_anchor;
```
bwesterb, line 643: You don't define the wire format `TrustAnchorIdentifier`.

bwesterb: Sorry, you do. If I understand correctly this is an ASN.1 oid and so needs byte-for-byte processing. Because of that it might be better to put after `proof_data` which is length prefixed?

bwesterb (approved):

draft-davidben-tls-merkle-tree-certs.md
davidben: Oh yeah, that's confusing. It's intended to be `TrustAnchorIdentifier` from here:
https://davidben.github.io/tls-trust-expressions/draft-beck-tls-trust-anchor-ids.html#name-overview

Probably we should just copy-paste the definition and write it as:
```
/* See Section 4.1 of draft-beck-tls-trust-anchor-ids */
opaque TrustAnchorIdentifier<1..2^8-1>;
```

bwesterb:
Bit of a nit, but why

```
struct {
    TrustAnchorIdentifier trust_anchor;
    opaque proof_data<0..2^16-1>;
} Proof;

struct {
    Assertion assertion;
    Proof proof;
} BikeshedCertificate;
~~~

and not directly

```
struct {
    Assertion assertion;
    TrustAnchorIdentifier trust_anchor;
    opaque proof_data<0..2^16-1>;
} BikeshedCertificate;
~~~

You could say the `proof_data` doesn't make sense without the `trust_anchor`, but the `proof_data` also doesn't make sense without the `assertion`.

davidben:
@bwesterb Somewhat looking forward to maybe doing something like https://github.com/davidben/tls-trust-expressions/issues/54, but yeah, let's inline it for now.

(I was OOO last week. Will take a pass to update this and look it over again.)

bwesterb:
> @bwesterb Somewhat looking forward to maybe doing something like [davidben/tls-trust-expressions#54](https://github.com/davidben/tls-trust-expressions/issues/54), but yeah, let's inline it for now.

I don't see how that'd work, but let's keep it as is then if you have something in mind.

davidben:
Sketched it out in https://github.com/davidben/merkle-tree-certs/issues/88

draft-davidben-tls-merkle-tree-certs.md
davidben: Done.

PR #90, opened 2024-09-12
Corrects test vectors based on a wrong length encoding of the DNS name

pohlm01:
See bwesterb/mtc#2 and bwesterb/mtc#4


davidben (approved):

davidben:
Thanks!

Issue #88, opened 2024-08-07
Multi-proof certificates


davidben:
(Not a fully-baked idea, just exploring the design space a bit.)

Should we support something like multi-proof certificates? The idea would be that the certificate looks something like:

```
struct {
    TrustAnchorIdentifier trust_anchor;
    opaque proof_data<0..2^16-1>;
} Proof;

struct {
    Assertion assertion;
    Proof proofs<0..2^16>;
} Certificate;
```

Then pair with something like https://github.com/davidben/tls-trust-expressions/issues/54 so that the server has a bunch of proofs available and the client can ask for some subset of them.

Things we could model this way:
* Signature-like proofs where you expect multiple independent entities to attest to it
* Something X.509 + CT-like where you have one signature from a domain validating entity and some signatures from logging entities
* MTCs with optional signatures over the tree head when the client isn't up-to-date

This also enables somewhat more compact representations on the subscriber. The assertion will contain a big PQ key and this way you just need the one blob in memory. Or if different clients' requests overlap, you only need to keep one copy of each individual component around.

One subtlety: if one proof depends on another proof, this structure flattens it and loses the information. That would need to be sufficiently encoded into the trust anchor identifiers and/or different credentials that you get a good combination.

davidben:
Some other subtleties:
* CertificatePropertyList applies to the whole Certificate. The properties get more complicated if they have to correlate with individual proofs
* Different proofs might expire at different times (e.g. MTCs have an upper bound). I'm thinking that, in #83, the expiry should be part of the Proof, with individual proofs being allowed to expire at different times

Other directions:
* Explode the structure in the protocol. First "certificate" in the TLS certificate list is the Assertion, the rest are individual Proofs
* Explode the structure in the subscriber / CA bit. Abandon the assertion size optimization and say you just get a bunch of assertion + single-proof tuples, stitch them back together by looking for matching assertions

bwesterb:
> One subtlety: if one proof depends on another proof, this structure flattens it and loses the information.

Which is the case for https://github.com/davidben/tls-trust-expressions/issues/54. However, we should think about what an analogue would look like when we have MTC in place.

I'd say the CA would create a signed _promise_ that it will include a certain assertion in a future batch. One can then gather cosignatures of witnesses that will check whether the CA indeed will include the assertion is the batch. The promise is akin to the regular certificate today, and the cosignature is like an SCT.

I can't quite make this fit neatly. Format of promise is alright.

```
struct {
[...]

Issue #41, opened 2023-06-05
Revocation


bwesterb:
> [[TODO: Is it worth defining an API for Merkle Tree CAs to publish a revocation list? That would allow automatically populating CRLite and CRLSets. Maybe that's a separate document.]]

bwesterb:
It's tempting to include revocations in the batches, by having either (Abridged)Assertions or Revocations as leafs. So we'd have, say:

```
enum {
    assertion(0),
    revocation(1),
    (2^8-1)
} LeafType;

struct {
    LeafType leaf_type;
    opaque leaf<0..2^24-1>;
} Leaf;

struct {
    uint32 batch_number;
    uint64 index;
} Revocation;
```

There are downsides:

- Revocation has the same delay as issuance. With `batch_duration`s of an hour this isn't too bad.
- We need to address revoking a revocation. I'd say revoking a revocation is a no-op.
- The revocations will take space in the Merkle Tree: proof sizes for a batch with a huge number of revocations will increase. (This could be offset by using a list of `Revocation`s as a leaf.)
-  It'll complicate the presentation of MTC: we can't say "we build a tree on top of assertions" anymore.

An alternative would be to have parallel batches/tree/signed windows where leafs are revocations instead of assertions. That only has the revocation delay as downside.

Finally, a bespoke HTTP interface could be created that would allow for quicker revocation. That'll force us to use a clock, which we don't really need so far (see #17.)

devonobrien:
I'm still not confident that building in CA-initiated revocation is the best or most preferable design for e.g. 10-14 day certificates. I'd rather include explicit certificate lifetime restrictions than integrate this kind of revocation (solutions like browser-made revocation lists can still be delivered out of band). At a policy level, the CABF currently appears poised to remove the need for revocation information certificates for <= 10 day X.509 certs, which aligns with the fastest reliable revocation given CRL and OCSP caching. 

Regarding your downsides
- We absolutely should not allow undoing revocation. That's not a currently permissible behavior in the WebPKI and we shouldn't build here, IMO.
- Even if every certificate in a batch were revoked, it should only increase proof size by ~one extra hash, right? We should already have a comfortable margin on proof sizes since we don't actually control issuance rates so they can vary greatly.

bwesterb:
> Even if every certificate in a batch were revoked, it should only increase proof size by ~one extra hash, right?

Yes. I was thinking of the case where you want to revoke million of certificates over multiple batches.

[...]

Issue #23, opened 2023-04-06
Evidence


bwesterb:
When a CA issues an assertion, it could provide evidence for it. For instance, a (list of) serialised dnssec proof(s) for a ACME dns-01 challenge or CAA record, if that was used. Most validations are http-01 for which no proof can be produced, except perhaps a signature of an independent multi-path DCV service, if that becomes a thing. Is evidence of assertions in scope? I suppose not for an initial version.

davidben:
Post-Photosynthesis triage: lots of things have changed, the general problem space still applies. Dunno if we specifically want it in this doc. (Initial testing phase is probably transitory enough to not need to be in this spec.)

davidben:
We ended up doing the funny bootstrap thing outside of the core MTC draft, since it's just a transitory, so I think we can close this for now.

PR #92, opened 2024-09-19
Add an explicit not_after time to the Proof

### LINKED ISSUE #83: Subscriber-imposed certificate expiry
Certificates currently all expire as the batch expires, which come from a technical requirement of the mechanism. However, when requesting a certificate, the subscriber may have even tighter opinions on expiry, based on its own key management goals. Probably not likely with MTC certificate lifetimes, but we do mean for BikeshedCertificate to be extensible and ACME does allow the subscriber to request a notAfter date.

If the subscriber requests a tighter notAfter date than the corresponding batch, the ACME server would need to choose between not issuing a Merkle tree certificate, or extending the subscriber's vulnerability to it key longer than requested. We could avoid that by adding an extra expiration time inside the certificate. The semantics would be that you check *both* that expiration time and the batch-level one. (Equivalently, require they be consistent. But checking both may be more straightforward since we need the RP to check anyway.)

Not sure whether this should go in Assertion + AbridgedAssertion, or MerkleTreeProofSHA256 + HashAssertionInput. I'm envisioning the Assertion being something that the subscriber puts together, that plugs into CSRs and any other proof-of-possession checks. Saying that the subscriber's Assertion is byte-for-byte identical to the one in the certificate gives a lot of nice properties, but in other proof types, a CA may wish to express a tighter lifetime than what the subscriber requested. That would suggest the CA-imposed expiry, in general, needs to be outside the assertion. (CA-imposed expiry is implicit in MTCs, but may not be in other constructions.) But then we might want PoPs to express a subscriber-imposed expiry, which would mean this "check both CA- and subscriber-imposed expiry separately" model needs to become universal.

Put another way, it's the difference between:
* The statement "the key 1234 is good for example.com" is true until 2024-01-31
* The statement "the key 1234 is good for example.com until 2024-01-20" is true until 2024-01-31

davidben:
This serves two purposes:

1. It allows a subscriber to easily know when the certificate is expired, without defining a separate certificate property.

2. It allows a subscriber to request a shorter expiration time than the batch would otherwise force. Fixes #83.

In looking at this, I realized that the document's organization is a bit all over the place. Maybe we should just pull the generic BikeshedCertificate parts out into their own section. But for now I'm leaving the overall structure as-is.


davidben:
Some notes and philosophizing:

**not_after vs batch expiration time** 

We could either have `not_after` be a constraint on top of what the batch would impose, with the relying party checking the minimum of the two, or say that `not_after` must be consistent with batch expiry, with the relying party rejecting the certificate if inconsistent.

Put another way, is `not_after` the subscriber-requested expiry, or the effective expiry?

I went with effective expiry because this means the subscriber does not need to know anything about the trust anchor to know when the certificate expires. Either it expires at the `not_after` time, or it was simply invalid and will be rejected at any time.

**Proof vs Assertion vs Certificate**

`not_after` could be part of the Assertion, part of the Proof, or just an extra field in the Certificate.

I think the litmus test for being part of the Assertion is "would you put this in a CSR?", and the answer is I think no. "example.com's key is 1234" is a fact about the world. A certificate is the CA making a point-in-time claim that the fact about the world is true.

Now, it being a point-in-time claim is kind of interesting. A CA that issues 14 day certificates and a CA that issues 1 year certificates do not actually perform different validation procedures. It's not like either CA has any better ability to make predictions about the future. So that might suggest that expiration times are all nonsense and maybe we should just say when the certificate was issued, with expiry being a relying party policy?

Except that's not only impractical but also doesn't capture some things. While expiry doesn't tell you anything about the validation process, it tells you how the whole ecosystem will manage this lifecycle of this certificate:

* Until a certificate expires, revocation for that certificate matters
* Until a certificate expires, subscribers should pay attention to it when monitoring transparency logs
* Until a certificate expires, subscribers can reasonably serve it in preference to a newer but larger certificate

Proof vs extra field in the Certificate is kind of meaningless right now. I think the litmus test is, if we end up doing #88, would it be a per-Proof or per-Certificate value? Because this is an effective expiry, I think the answer is per-Proof. Different MTC CAs may force different `not_after` dates, etc.

davidben:
Another random thought / justification

**Opaque vs part of framework**

We could either make it part of the BikeshedCertificate framework, or make it something that you put in the specific proof type. Ultimately these are mostly isomorphic. If something is specific to the proof type, subscribers can't read it generically, but then we can define a certificate property that they _do_ know how to read.

But since subscribers should always be able to read this, and these kinds of offline assertions are inherently related to time, putting it in the framework made sense.

bwesterb:
Some more things to ponder:

1. How does a subscriber request the earlier expiration? Previously, the client would just send an assertion as assertion-publishing-request. Now we need a `AssertionSigningRequest` different from an Assertion that contains the expiry.

2. Do we need a timestamp or can we use batches are granularity?

3. Maybe also good to ponder revocation again. Depending on how we solve that, we might achieve what we want, by issuing and then immediately revoking for a set time. #41

bwesterb:
There is an issue: if you don't store not_after in the abridgedassertion, then the trust service can't compute HashAssertion and thus verify the tree.

davidben:
Ah nice catch! I forgot about the HTTP interface.

AbridgedAssertionAndTimestamp is a mouthful. It is annoying that probably every proof type will need a family of "[abridged] assertion and not after" structures and we don't good a word for that. Should it actually go in the assertion?? Trouble is this makes it harder to staple multiple proofs a la #88 if the proofs happen not to be clocked similarly, or have other lifetime constraints, and I kinda like the freedom to build PKIs out of that.

draft-davidben-tls-merkle-tree-certs.md
```diff
@@ -496,9 +496,9 @@ The CA exposes all of this information in an HTTP {{!RFC9110}} interface describ
 
 ## Issuance Queue and Scheduling
 
-The CA additionally maintains an issuance queue, not exposed via the HTTP interface.
+The CA additionally maintains an issuance queue, not exposed via the HTTP interface. Each element of the issuance queue is an assertion to be certified, and an optional requested expiration time. The final `not_after` will be computed from both this requested time and the expiration time of the whole batch.
```
lukevalenta, line 499: In our current implementation (which of course we're happy to change), subscribers submit an `AssertionRequest` to the CA which contains the assertion, evidence, and requested maximum `not_after`. But before queuing that `AssertionRequest`, the CA computes the effective `not_after` as the minimum of the subscriber-requested `not_after`, the next-to-be-issued-batch's expiration time, and any constraints for the evidence (e.g., X.509 cert or revocation check expiry).

The reason we compute the effective `not_after` date at queue time is that otherwise we'd have to re-validate all of the evidence at issuance time, which could be a bottleneck. It also means that we use the next-to-be-issued batch's expiration time instead of the actual batch's expiration time, which prevents assertions from having a longer-than-expected lifetime if there are issuance delays (related to #105).

lukevalenta: Regarding evidence (#23), that's also going into the queue in our implementation as part of the `AssertionRequest`, but should be fine to leave that out of this PR.

davidben: Ah hmm. I mean, both behaviors seem perfectly legitimate to me, so ideally the text would admit both. I guess what you're doing could be modeled as computing a different requested expiration time _before_ entering the queue, which happens to be always less than `batch_not_after`. But that is definitely on the confusing side. Will wordsmith a bit.

davidben:
Alright, rebased and then reworked a bunch. This probably warrants a fresh review and possibly looking at individual commits for clarity.

I rebased the PR.

I addressed @lukevalenta's comment by being less prescriptive about where you get the expiration time.

I addressed @bwesterb's comment by putting not_after into AbridgedAssertion.

In doing that, I realized calling this an abridged assertion gets... kinda weird, so I did a bigger change in hopes of resolving this. I tried calling them "batch entries" which seems to work? This resolves a few issues:
* Abridged assertions contain more information than just the assertion (maybe not_after should just be in the assertion??? https://github.com/davidben/merkle-tree-certs/pull/92#issuecomment-2362330928 was the rationale for sticking them with the proof)
* Abridged assertions are currently presented as a funny implementation detail of leaf hashing, but they're more than that. They're also the thing we serve in the HTTP list.
* So the phrasing around "batch entries" tries to talk about them as being *the* contents of the Merkle Tree.

Finally, I noticed a weird inconsistency in how we refer to hash values so cleaned that up.

Thoughts?

bwesterb:
> So the phrasing around "batch entries" tries to talk about them as being _the_ contents of the Merkle Tree.

What about *commitments*? Or is that too crypto-y?

lukevalenta (approved):
Modulo a missing period and whether to call the leaves 'commitments' or 'entries' (I think either is fine, although 'entries' is more succinct), looks good to me!

One implementation consideration (which doesn't need to be part of the spec): right now the [Key()](https://github.com/bwesterb/mtc/blob/main/mtc.go#L1285) function computes a key for an abridged assertion, which is used in various indices and by authenticating parties to request a BikeshedCertificate for a given issued assertion. I think we'll want to omit the `NotAfter` field from that key it's mutable. Instead, the key should be derived from fields in the Assertion structure itself.

draft-davidben-tls-merkle-tree-certs.md
```diff
@@ -475,45 +475,76 @@ The CA also maintains a latest batch number, which is the number of the last bat
...
+~~~
+/* In this document, hash is always SHA-256, and hash.length is 32. */
+opaque HashValue[hash.length];
+
+struct {
+    SubjectType subject_type;
+    HashValue subject_info_hash;
+    Claim claims<0..2^16-1>;
+    Timestamp not_after;
+} MerkleTreeEntry;
```
bwesterb, line 499: Not ideal to have *batch entry* and *merkle tree entry*. We should make this `BatchEntry` or use *Merkle tree entry* everywhere.

bwesterb (approved):
Don't think naming is ideal yet, but let's just move forward.

draft-davidben-tls-merkle-tree-certs.md
davidben: Ah yeah, done.

davidben:
> One implementation consideration (which doesn't need to be part of the spec): right now the [Key()](https://github.com/bwesterb/mtc/blob/main/mtc.go#L1285) function computes a key for an abridged assertion, which is used in various indices and by authenticating parties to request a BikeshedCertificate for a given issued assertion. I think we'll want to omit the NotAfter field from that key as it's mutable. Instead, the key should be derived from fields in the Assertion structure itself.

Ah yeah, another reason to have not_after be part of the Proof rather than the Assertion I guess.

davidben:
> What about commitments? Or is that too crypto-y?

I feel like "entry" feels more understandable than "commitment" to me, but I'm perhaps not up on the relevant lingo here. I was just thinking "this is what is actually in the log", because the log does not (necessarily) log the unhashed keys.

*shrug* We can always reshuffle this later.

PR #102, opened 2025-02-13
Rework discussion of the transparency ecosystem

### LINKED ISSUE #99: Rename transparency service?
The transparency service is described as a single service just for clarity, but its role can be fulfilled by whatever the relying party wants, including e.g. a quorum of multiple TSes or whatever else.

This has always been the intent of this design, but it seems folks get confused by this. Perhaps we can rename it to avoid those folks getting confused.

Also "transparency service" makes it sound more lofty than it really is. It's basically just a mirror (or set of mirrors) of what the CA publishes, trusted to remain accessible and not introduce split views. Maybe "trusted mirror(s)"?

davidben:
Although the transparency service was clearly described throughout the document as an abstraction and not actually necessarily one single centralized service, this seems to have been a point of confusion. The intent was to observe that this system actually allows the ecosystem to freely design trust models, mirroring roles, etc., all without impact to interoperability. As this all happens out of band, neither certificate issuance nor certificate usage is impacted by these choices, except that a certificate may or may not be selected by negotiation.

This seems to have been a confusing notion. To avoid that mess, recut these concepts in the document.

First, define the role of a mirror as a concrete service with some responsibility. This picks up the mirroring aspects of the old example transparency services, but not the direct connection to an RP or trustworthiness. It's a freefloating service with some responsibility. You may or may not believe it meets those responsibilities.

Then, define a relying party policy, for how an RP gets gets validity windows. This is more explicitly the source of implementation dependence. RPs can choose to trust some combination of mirrors when formulating this policy. We give several example policies, but don't mandate anything in particular.

Rewrite the rest of this text according to this framing.

Closes #99.

### DIFF
--- a/draft-davidben-tls-merkle-tree-certs.md
+++ b/draft-davidben-tls-merkle-tree-certs.md
@@ -807,7 +807,7 @@ Transparency:
 
 How the relying party does this has no direct impact on certificate issuance ({{issuing-certificates}}) or usage ({{using}}). Different relying parties can obtain validity windows from different sources or apply different policies, while supporting the same unmodified certificates and CAs. Thus this section does not prescribe particular policies or mechanisms, but provides examples and general guidance. Relying parties MAY implement policies other than those described below, and MAY incorporate entities acting in roles not described in this document.
 
-Relying parties SHOULD ensure transparency by obtaining validity windows from the CA and/or some combination of trusted mirrors. The relying party picks sources which, together, are trusted to satisfy the requirements described in {{mirrors}}. Mirrors allow the relying party to maintain transparency in the face of a misbehaving CA that may, for example, stop serving some unauthorized certificate in a batch to evade detection.
+Relying parties SHOULD ensure transparency by obtaining validity windows from the CA and/or some combination of trusted mirrors. The relying party picks sources which, together, are trusted to satisfy the requirements described in {{mirrors}}. Mirrors allow the relying party to maintain transparency in the face of a misbehaving or compromised CA that may, for example, stop serving some unauthorized certificate in a batch to evade detection.
 
 A relying party might trust a combination of mirrors, and only accept a validity window once some minimum number of mirrors have consumed it. When combining multiple mirrors, the following procedure determines the latest validity window for an update:
 


davidben:
Please feel free to go over this with a fine-toothed comb and nitpick. This probably does need a editing pass. (I'll try to do a review pass myself later in the day.)

draft-davidben-tls-merkle-tree-certs.md
```diff
@@ -280,9 +280,9 @@ First, the CA issues a certificate to the authenticating party:
 Next, tree heads flow to the relying party, after being durably and consistently logged. This occurs periodically in the background, unconnected from uses of individual certificates:
 
 {:start="4"}
-4. The transparency service periodically downloads the abridged assertions, recreates the Merkle Tree, and validates the window signature. It mirrors them for monitors to observe. See {{transparency-service}}.
+4. Mirrors periodically download the abridged assertions, recreate the Merkle Tree, and validate the window signature. They mirror the contents for monitors to observe. See {{transparency-ecosystem}}.
 
-5. The relying party periodically fetches the latest validity window from the transparency service. This validity window will contain the new tree head.
+5. The relying party periodically fetches the latest validity window from mirrors that it trusts. This validity window will contain the new tree head.
```
lukevalenta, line 285: I think we're missing the distinction between the RP and the RP's out-of-band update service here. This is important because of the privacy implications of a browser reaching out directly to mirrors, for example. Maybe we need separate boxes for those two entities to make things crystal clear?
```suggestion
5. The relying party's update service periodically fetches the latest validity window from mirrors that it trusts. This validity window will contain the new tree head.
```

draft-davidben-tls-merkle-tree-certs.md
```diff
@@ -790,57 +793,47 @@ The transparency service maintains a mirror of the CA's latest batch number, and
...
 
-7. Verify the validity window signature, as described in {{signing}}. If the signature is invalid, abort this procedure with an error.
+A relying party might trust a combination of mirrors, and only accept a validity window once some minimum number of mirrors have consumed it. Such a relying party would fetch the latest batch number from each mirror, then compute the highest batch number that is contained in sufficiently many mirrors. It would then fetch that validity window from each containing mirror, check their tree hashes match and, if so, accept the update.
 
-8. If the old and new validity windows contain overlapping batch numbers, verify that the tree hashes match. If not, abort this procedure with an error.
+This procedure MAY also be performed by the relying party's update service, or some aggregation service, which forwards the signatures from each mirror to the relying party. This allows the relying party to avoid directly contacting each mirror. [[TODO: This assumes mirrors re-sign validity windows, but we have not actually defined that yet. #101]]
```
lukevalenta, line 816: Just noting that if we go this route, it would be nice to leverage the co-signing mechanism of [tlog-witness](https://github.com/C2SP/C2SP/blob/main/tlog-witness.md).

lukevalenta (approved):
Left some suggestions, but overall the change looks good to me.

draft-davidben-tls-merkle-tree-certs.md
```diff
@@ -236,11 +236,11 @@ Merkle Tree certification authority (CA):
 Relying party:
 : The party whom the authenticating party presents its identity to. In TLS, this is the side receiving the Certificate and CertificateVerify message.
```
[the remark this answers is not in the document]
davidben: Not sure. Rather than be too prescriptive about the structure of the RP, I tried making the text outside policy bits more general, and then reworded the policy bits a bit.

Still not sure if it's quite right, but maybe an okay direction. I'm sure we'll keep refining this more (and figuring out what to specify and what not) later, so I was mostly aiming here to contain the fluid part to the Transparency Ecosystem section.

WDYT?

draft-davidben-tls-merkle-tree-certs.md
```diff
@@ -790,57 +793,55 @@ The transparency service maintains a mirror of the CA's latest batch number, and
...
+How the relying party does this has no direct impact on certificate issuance ({{issuing-certificates}}) or usage ({{using}}). Different relying parties can obtain validity windows from different sources or apply different policies, while supporting the same unmodified certificates and CAs. Thus this section does not prescribe particular policies or mechanisms, but provides examples and general guidance. Relying parties MAY implement policies other than those described below, and MAY incorporate entities acting in roles not described in this document.
 
-3. If `new_latest_batch` equals `old_latest_batch`, finish this procedure without reporting an error.
+Relying parties SHOULD ensure transparency by obtaining validity windows from the CA and/or some combination of trusted mirrors. The relying party picks sources which, together, are trusted to satisfy the requirements described in {{mirrors}}. Mirrors allow the relying party to maintain transparency in the face of a misbehaving CA that may, for example, stop serving some unauthorized certificate in a batch to evade detection.
```
bwesterb, line 810: I like *misbehaving or compromised*. Transparency isn't just about keeping CAs honest, but also giving a well-behaving CA some assurance they've not been compromised.

bwesterb (approved):

draft-davidben-tls-merkle-tree-certs.md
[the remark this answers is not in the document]
lukevalenta: Yeah, I like the approach you took here 👍

draft-davidben-tls-merkle-tree-certs.md
lukevalenta: The change to de-emphasize the RP directly contacting the mirror looks good!

draft-davidben-tls-merkle-tree-certs.md
```diff
@@ -790,57 +793,55 @@ The transparency service maintains a mirror of the CA's latest batch number, and
...
 
-4. If `new_latest_batch` is less than `old_latest_batch`, abort this procedure with an error.
+A relying party might trust a combination of mirrors, and only accept a validity window once some minimum number of mirrors have consumed it. When combining multiple mirrors, the following procedure determines the latest validity window for an update:
 
-5. If the issuance time for batch `new_latest_batch` is after the current time (see {{parameters}}), abort this procedure with an error.
+1. Fetch the latest batch number from each mirror.
 
-6. Fetch the validity window with `new_latest_batch` from each mirror that returned an equal or higher latest batch number. If any fetches fail, or if the results do not match across all mirrors, abort this procedure with an error.
+2. Compute the highest batch number that satisfies the policy. For example, if requiring windows be represented in at least two mirrors, use the second highest batch number.
```
lukevalenta, line 816: nit: if the two mirrors both serve the highest batch number then we don't want the second-highest
```suggestion
2. Compute the highest batch number that satisfies the policy. For example, if requiring windows be represented in at least two mirrors, use the second-to-last batch number after sorting in ascending order.
```

draft-davidben-tls-merkle-tree-certs.md
davidben: Fair enough. I guess I would say that the CA is the service with a private key and the _CA operator_ is the entity that maintains it. If the CA is compromised, the _CA operator_ may well be honest, but the service is still misbehaving. But people are often sloppy enough about this that adding the extra word SGTM.

```suggestion
Relying parties SHOULD ensure transparency by obtaining validity windows from the CA and/or some combination of trusted mirrors. The relying party picks sources which, together, are trusted to satisfy the requirements described in {{mirrors}}. Mirrors allow the relying party to maintain transparency in the face of a misbehaving or compromised CA that may, for example, stop serving some unauthorized certificate in a batch to evade detection.
```

PR #108, opened 2025-03-13
Fix mirroring procedure issues in #107

### LINKED ISSUE #107: Mirroring nits
1. §7.1.1 (2)

> Let new_latest_batch be the result and old_latest_batch be the currently mirrored value. If new_latest_batch equals old_latest_batch, finish this procedure without reporting an error.

Maybe just define `old_latest_batch` to be -1 if no batch was mirrored yet.

2. We should mention that the mirror checks the batch number in the signed validity window.

davidben:
Closes #107

### DIFF
--- a/draft-davidben-tls-merkle-tree-certs.md
+++ b/draft-davidben-tls-merkle-tree-certs.md
@@ -800,7 +800,7 @@ A mirror maintains a copy of the CA's latest batch number, and batch state. Roug
 
 1. Fetch the latest batch number.  If this fetch fails, abort this procedure with an error.
 
-2. Let `new_latest_batch` be the result and `old_latest_batch` be the currently mirrored value. If `new_latest_batch` equals `old_latest_batch`, finish this procedure without reporting an error.
+2. Let `new_latest_batch` be the result. Let `old_latest_batch` be the currently mirrored value, or -1 if no batch was yet mirrored. If `new_latest_batch` equals `old_latest_batch`, finish this procedure without reporting an error.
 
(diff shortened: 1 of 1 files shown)


draft-davidben-tls-merkle-tree-certs.md
```diff
@@ -816,9 +816,11 @@ A mirror maintains a copy of the CA's latest batch number, and batch state. Roug
 
    2. Compute the tree head for the assertion list, as described in {{building-tree}}. If this value does not match the fetched tree head, abort this procedure with an error.
 
-   3. Compute the ValidityWindow structure and verify the signature, as described in {{signing}}. Set `tree_heads[0]` to the tree head fetched above. Set the other values in `tree_heads` to the previously mirrored values. If signature verification fails, abort this procedure with an error.
+   3. Compute the expected ValidityWindow structure as follows: Set `batch_number` to `i`. Set `tree_heads[0]` to the tree head fetched above. Set the other values in `tree_heads` to the previously mirrored values.
```
davidben, line 819: I noticed that this procedure is written as if you construct the ValidityWindow you expect rather than fetching it. I kept that but depending on how the HTTP format shakes out, maybe that'll change.

(Also I'm getting convinced that maybe we should just chain the hashes together. It makes describing signatures more straightforward because you don't need to sign over the whole window, just the last hash. Though we still need to talk about validity windows because clients need to receive the whole hash.)

bwesterb:
This doesn't address the `old_latest_batch` being undefined yet.

bwesterb (approved):

PR #110, opened 2025-03-13
Downgrade hash agility to a NOTE

cjpatton:
The draft suggests hash functions other than SHA-256 might be used for the Merkle tree, but doesn't specify how. I think the intention is to signal to the reader that there's room for adding this agility in the future, so say so in an author note.

### DIFF
--- a/draft-davidben-tls-merkle-tree-certs.md
+++ b/draft-davidben-tls-merkle-tree-certs.md
@@ -417,7 +417,7 @@ This section describes the structure of Merkle Tree certificates and defines the
 A Merkle Tree certification authority is defined by the following values:
 
 `hash`:
-: A cryptographic hash function. In this document, the hash function is always SHA-256 {{SHS}}, but others may be defined.
+: A cryptographic hash function. In this document, the hash function is always SHA-256 {{SHS}}. [[NOTE: We can add support for others as needed, though the CA would need to indicate which hash function it uses.]]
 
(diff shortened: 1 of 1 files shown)


davidben:
Oh this is slightly interesting. Arguably, by what's in the draft right now, we actually have all the hash agility mechanisms we need. The hash is one of the parameters of the CA. If the relying party knows the CA, it's assumed to know all the CA parameters, which includes the the hash. So based on what's in the document _right now_, the only way we assume SHA-256 is just we vaguely hint that people should do that.

But this is kind of cheating because we said how to represent a trust anchor for relying parties. Once we do that, we'll need to start baking in the hash or adding an enum field with only one defined integer, or whatever. So... I think probably the thing to do is to get that done and then write whatever makes sense based on what we do there.

cjpatton:
SGTM. Another thing we might need to worry about is the output size of the hash, which is hard-coded to 32 in a lot of places. I don't think this will matter much in practice though, so we could explicitly restrict the set of hash functions with which MTCs are compatible.

cjpatton:
The only thing that stuck out to me was: 
```
    uint8 label[32] = "Merkle Tree Crts ValidityWindow\0";
```
I'm not sure if you picked 32 because you wanted this to match the ouptput size of SHA256. But I don't think it really matters :)

davidben:
Oh hah! I think that was just a coincidence. :-D

PR #117, opened 2025-07-03
Pick a concrete construction for ACME

davidben:
Based on discussions with @aarongable (this look good?)


draft-davidben-tls-merkle-tree-certs.md
```diff
@@ -1013,17 +1013,15 @@ Authenticating parties SHOULD preferentially use signatureless certificates over
 
 # ACME Extensions
 
-[[TODO: This section hasn't been written yet. Instead, what follows is an informal discussion. #13 ]]
+This section describes how to issue Merkle Tree certificates using ACME {{!RFC8555}}.
 
-{{Section 6 of !I-D.ietf-tls-trust-anchor-ids}} defines how to get trust anchor ID metadata for a certificate.
+To communicate trust anchor IDs for use with TLS (see {{use-in-tls}}), ACME servers SHOULD send the "application/pem-certificate-chain-with-properties" content type, as described in {{Section 6 of !I-D.ietf-tls-trust-anchor-ids}}. ACME clients supporting Merkle Tree certificates SHOULD implement this extension.
```
aarongable, line 1018: nit: I think this paragraph could be more explicit -- clients which support merkle tree certificates should include "application/pem-certificate-chain-with-properties" in their Accept header, and servers which provide trust anchor IDs should honor that Accept header and send that content type when they see it.

PR #120, opened 2025-08-11
Fix typo in definition of k for subtree proofs

lukevalenta:


draft-davidben-tls-merkle-tree-certs.md
```diff
@@ -321,7 +321,7 @@ Given an ordered list of `n` inputs, `D_n = {d[0], d[1], ..., d[n-1]}`, {{Sectio
 A *subtree* of this Merkle Tree is defined by two integers, `start` and `end`, such that:
 
 *  `0 <= start < end <= n`
-* If `k` is the largest power of 2 that is greater than or equal to `end - start`, `start` must be a multiple of `k`
+* If `k` is the largest power of 2 that is less than or equal to `end - start`, `start` must be a multiple of `k`
```
davidben, line 324: Hmm. I think greater is correct, unless I'm confused? `[2, 5)` is not a valid subtree because it makes this tree:

```
  /\
 /\ |
2 3 4
```

But then `[0, 5)` looks like this, which doesn't contain that tree:

```
     /\
    /  \
   /\   |
  /  \  |
 /\  /\ |
0 1 2 3 4
```

So if the subtree has size 3, we need start to be 4-aligned, not just 2-aligned.

But in that case "largest power of 2" should be "smallest power of 2".

lukevalenta: Ah yes, you're right--thanks for the catch and the counterexample! I agree the correct version is "If `k` is the smallest power of 2 that is greater than or equal to `end - start`, `start` must be a multiple of `k`", since for any size subtree we want the start index to be aligned with the next-greatest power of two.

I got confused since [RFC9162](https://datatracker.ietf.org/doc/html/rfc9162#section-2.1.1) and [the subtree consistency proof section](https://www.ietf.org/archive/id/draft-davidben-tls-merkle-tree-certs-06.html#section-4.3.1) say "let k be the largest power of two smaller than n", but we're dealing with a different `k` here--that `k` is for a split index, but the `k` here is a restriction on the subtree start index. Maybe we pick a new letter like `s` for "start"?

davidben (approved):

PR #121, opened 2025-08-12
Copy-edit of draft charter

aarongable:
Incorporate minor edits from my own copy-level review of the document.

### DIFF
--- a/charter-ietf-plants.md
+++ b/charter-ietf-plants.md
@@ -12,7 +12,7 @@ Overhead from post-quantum signatures and keys will add significant costs in two
 
 * Each log entry contains an entire certificate, with public key and signature. Post-quantum overhead is multiplied across every entry, increasing the costs to log operators and the rest of the transparency ecosystem.
 
-* Relying parties are presented with signatures from the CA and CT logs. Post-quantum overhead is multiplied per signature, increasing the latency of the TLS handshake.
+* Relying parties are presented with signatures from the CA and CT logs. Post-quantum overhead is multiplied per signature, increasing the size and latency of the TLS handshake.
 
 The PLANTS Working Group will define a mechanism that integrates log construction into certificate issuance. This enables techniques where signature costs can be amortized over multiple key/identifier bindings, e.g. by signing Merkle Tree hashes.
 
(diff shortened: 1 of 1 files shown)


charter-ietf-plants.md
```diff
@@ -12,21 +12,21 @@ Overhead from post-quantum signatures and keys will add significant costs in two
...
 
-3. How the certificate constructions may be provisioned with ACME and used in TLS.
+3. How the certificate constructions may be provisioned with ACME (RFC 8555) and used in TLS.
 
-As part of this work, the Working Group may extend X.509, e.g. with new extensions or signature algorithms. As appropriate, the PLANTS Working Group will liaise with the LAMPS Working Group to ensure adequate lighting for this work and help it grow. As needed, the Working Group may also define extensions to ACME and TLS to integrate its certificate constructions. In doing so, it is expected to liaise with the TLS and ACME Working Groups for cross-pollination.
+As part of this work, the Working Group may extend the RFC 5280 profile of X.509, e.g. with new extensions or signature algorithms. As appropriate, the PLANTS Working Group will liaise with the LAMPS Working Group to ensure adequate lighting for this work and help it grow. As needed, the Working Group may also define extensions to ACME and TLS to integrate its certificate constructions. In doing so, it is expected to liaise with the TLS and ACME Working Groups for cross-pollination.
```
aarongable, line 27: Reasoning: because X.509 isn't maintained by the IETF, but RFC 5280 is.

charter-ietf-plants.md
```diff
@@ -12,21 +12,21 @@ Overhead from post-quantum signatures and keys will add significant costs in two
...
 
-Though not the initial focus, the PLANTS Working Group may consider other properties of transparent PKIs to improve upon the status quo, such as auditing, monitoring, or revocation. If concrete, feasible improvements are identified, the Working Group may recharter to seed new, secondary deliverables that build on its initial work.
+Though not the initial focus, the PLANTS Working Group may consider other properties of transparent PKIs to improve upon the status quo, such as auditing, monitoring, or revocation. If feasible concrete improvements are identified, the Working Group may recharter to seed secondary deliverables that build on its initial work.
```
aarongable, line 29: Reasoning: just to reduce the number of commas in the sentence, to ease reading.

bwesterb (approved):

charter-ietf-plants.md
davidben: Ah hmm, I'm very unclear on the right semantics here. Is defining a new extension or sigalg extending the RFC 5280 profile? RFC 8410 doesn't update RFC 5280 or anything. It just defines a new sigalg that could be used with any X.509-based-system, RFC 5280 or otherwise.

I guess there's an ambiguity whether "extending RFC 5280" means defining new stuff one can use with RFC 5280, or update some of the text in the RFC 5280 profile. I was mostly thinking the former, though I suppose we might end up needing to do that. x509-alg-none needed to update some of it.

In that same vein, extending X.509 could mean defining new stuff one can use with X.509, or overriding X.509's text, or bumping the version number and all. We definitely shouldn't do that last one; at that point, there's no benefit in keeping continuity with X.509.

For some reason "extending the ... profile of ...." reads more to me like updating the profile itself, while "extending ..." could be read as either.

We could say "the WG may define new X.509 extensions and signature algorithms" but then we need to enumerate all the extension points. I looked at the old CURDLE and LAMPS charters and they all say PKIX, so perhaps that's the way around this. I wonder if we should be saying PKIX instead of X.509 in general or something...

davidben: Perhaps simply:

```suggestion
As part of this work, the Working Group may extend PKIX (RFC 5280), e.g. with new extensions or signature algorithms. As appropriate, the PLANTS Working Group will liaise with the LAMPS Working Group to ensure adequate lighting for this work and help it grow. As needed, the Working Group may also define extensions to ACME and TLS to integrate its certificate constructions. In doing so, it is expected to liaise with the TLS and ACME Working Groups for cross-pollination.
```

PR #122, opened 2025-08-13
X509 is not a specification we own

mcr:
X.509 is not even what we use, it's RFC5280.
The ITU-T should get *NO* credit here. They have never helped.


bwesterb (approved):

davidben:
Thanks! @aarongable beat you to it in https://github.com/davidben/merkle-tree-certs/pull/121, so closing this as already applied. :-)

PR #123, opened 2025-08-13
reword goals

mcr:


charter-ietf-plants.md
```diff
@@ -4,7 +4,7 @@ This is a draft charter for a new IETF working group. The working group does not
 
 ## PKI, Logs, And Tree Signatures (PLANTS)
 
-The goal of the PLANTS Working Group is to trim the costs of large post-quantum signatures on PKIs with Certificate Transparency (CT; RFC 6962 and RFC 9162), when used in interactive protocols like TLS (RFC 8446).
+The goal of the PLANTS Working Group is to rethink the tradeoffs between certificate signatures and Transparency Logs.  This is driven the size and performance of quantum-safe algorithms.  The initial focus of the WG is on PKIs that use Certificate Transparency (CT; RFC 6962 and RFC 9162), with interactive protocols like TLS (RFC 8446) and IKEv2 (RFC7296)
```
bwesterb, line 7: This "tradeoff" doesn't type-check for me, do you perhaps mean trading transparency against performance? Or did you mean "rethink division of tasks between certificate authorities and transparency logs"?

mcr: okay, so maybe some words are missing here.
certificates cost bytes and CPU.
transparency logs cost bytes and CPU.
Quantum-safe algorithms change the values, and thus the tradeoffs.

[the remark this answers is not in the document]
davidben: And if the IKEv2 community is ready to build out a transparent PKI while this work is ongoing, there's nothing stopping the WG from discussing other use cases then. But I don't think the charter should commit us to doing something for IKEv2 prematurely.

davidben:
Closing since the charter is being tracked elsewhere now.

PR #133, opened 2025-08-22
charter: slight tweak to first deliverable suggested by Dennis

bwesterb:
https://mailarchive.ietf.org/arch/msg/plants/4bdDJLDkQfsM7qiW5lW2PjBwKZY/

### DIFF
--- a/charter-ietf-plants.md
+++ b/charter-ietf-plants.md
@@ -18,7 +18,7 @@ The PLANTS Working Group will define a mechanism that integrates log constructio
 
 The Working Group will initially put down roots and define the mechanisms needed to interoperably construct and consume certificates:
 
-1. A transparency log structure, maintained by a CA, containing the key/identifier bindings that the CA has certified.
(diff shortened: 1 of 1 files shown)


charter-ietf-plants.md
```diff
@@ -18,7 +18,7 @@ The PLANTS Working Group will define a mechanism that integrates log constructio
 
 The Working Group will initially put down roots and define the mechanisms needed to interoperably construct and consume certificates:
 
-1. A transparency log structure, maintained by a CA, containing the key/identifier bindings that the CA has certified.
+1. An externally monitorable transparency log structure, maintained by a CA, containing the key/identifier bindings that the CA has certified.
```
davidben, line 21: This prompts a few unanswered questions, but I think that's a good thing (i.e. let's leave them unanswered for the WG):
* External to whom? I think the obvious answer is "the public can monitor", but private PKIs might have a more nuanced answer, so leaving that to the deployment seems solid 
* How is it accessed? Is it the CA that guarantees access or mirrors or both? Does the spec have an opinion on this or do we leave it to deployments?

PR #143, opened 2025-09-10
Override issuer encoding used in early implementations

### LINKED ISSUE #142: X.509 name construction isn't compatible with Go
@cjpatton noticed that Go's X.509 parser doesn't actually accept the name construction we picked: https://github.com/golang/go/issues/75260

Might be worth picking another one. Looks like Go currently only supports the following attribute values:
* T61String
* PrintableString
* UTF8String
* BMPString
* IA5String
* NumericString

Compatibility-wise, it doesn't particularly matter for Go _relying parties_ because old relying parties won't accept these anyway. But one of the benefits of living with X.509 is that existing authenticating parties can still use these. Authenticating parties broadly only need to parse enough of the certificate to get it through their configuration pipelines and into the TLS stack. (And extract a few things like the public key.)

Particularly to help early experiments, before the Go issue is fixed, it might be worth doing this. Some options off-hand:

1. Don't constrain the name. Just say that a log has both an X.509 name and a trust anchor ID, and everyone needs to remember they're associated. This makes configuration more complicated.
2. Use the text form of the trust anchor ID and stick it in a UTF8String. Burns a couple more bytes, but more straightforwardly compatible.
3. Turn the trust anchor ID into an (absolute) OID. Use that as the attribute type and `UTF8String("")` as the value.

cjpatton:
Closes #142.

RELATIVE-OID isn't yet widely implemented. For example, see <https://github.com/golang/go/issues/75260>. To unblock 
early deployments, we'll go with UTF8String instead.

### DIFF
--- a/draft-davidben-tls-merkle-tree-certs.md
+++ b/draft-davidben-tls-merkle-tree-certs.md
@@ -645,7 +645,17 @@ The attribute's value is a RELATIVE-OID containing the trust anchor ID's ASN.1 r
 1.3.6.1.5.5.7.25.TBD=#0d0481fd5901
 ~~~
 
-For initial experimentation, early implementations of this design will use the OID 1.3.6.1.4.1.44363.47.1 instead of `id-rdna-trustAnchorID`.
+For initial experimentation, early implementations of this design will:
+
+1. Use UTF8String to represent the attribute's value rather than RELATIVE-OID. The UTF8String contains trust anchor ID's ASCII representation, e.g. `324731.1`.
+
+1. Use the OID 1.3.6.1.4.1.44363.47.1 instead of `id-rdna-trustAnchorID`.
+
+For example, the distinguished name for a log named `32473.1` would be represented in syntax of {{?RFC4514}} as:
+
+~~~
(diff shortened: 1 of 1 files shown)


draft-davidben-tls-merkle-tree-certs.md
```diff
@@ -645,7 +645,17 @@ The attribute's value is a RELATIVE-OID containing the trust anchor ID's ASN.1 r
 1.3.6.1.5.5.7.25.TBD=#0d0481fd5901
 ~~~
 
-For initial experimentation, early implementations of this design will use the OID 1.3.6.1.4.1.44363.47.1 instead of `id-rdna-trustAnchorID`.
+For initial experimentation, early implementations of this design will:
+
+1. Use UTF88String to represent the attribute's value rather than RELATIVE-OID.
```
davidben, line 650:
```suggestion
1. Use UTF8String to represent the attribute's value rather than RELATIVE-OID. The UTF8String contains trust anchor ID's ASCII representation, e.g. `324731.1`.
```

davidben (approved):

PR #145, opened 2025-09-17
Clarify loop exit condition in subtree inclusion proof verification

lukevalenta:
Entering the loop, either `LSB(fn)` is already set, or `fn` is equal to `sn` which is non-zero. Thus, the right-shifting loop will only ever exit due to `LSB(fn)` being set.


davidben (approved):
I considered suggesting we add an explanatory sentence to “This is the same as the procedure […]” because it’s now not as obvious, but it seemed awkward so… meh.

PR #149, opened 2025-10-06
Unify the two branches in the consistency proof check

davidben:
Also replace prepending to the proof list with removing in the other branch. It makes it a little obvious how to implement this without making a copy of the proof. This should also hopefully make the relationship between inclusion and consistency proof evaluation clearer.

While I'm here, use '1.' for the steps so we don't have to renumber it as it changes.

(CC @nharper)

### DIFF
--- a/draft-davidben-tls-merkle-tree-certs.md
+++ b/draft-davidben-tls-merkle-tree-certs.md
@@ -516,32 +516,25 @@ The following procedure can be used to verify a subtree consistency proof.
 
 Given a Merkle Tree over `n` elements, a subtree defined by `[start, end)`, a consistency proof `proof`, a subtree hash `node_hash`, and a root hash `root_hash`:
 
-1. If `end` is `n`, run the following:
-   1. Set `fn` to `start` and `sn` to `end - 1`.
-   2. Set `r` to `node_hash`.
-   3. Until `LSB(fn)` is set or `sn` is `0`, right-shift `fn` and `sn` equally.
-   4. For each value `p` in the `proof` array:
-      1. If `sn` is `0`, then stop iteration and fail the proof verification.
-      2. Set `r` to `HASH(0x01 || p || r)`.
-      3. Until `LSB(sn)` is set, right-shift `sn`.
-      4. Right-shift `sn` once more.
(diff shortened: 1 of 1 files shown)


draft-davidben-tls-merkle-tree-certs.md
```diff
@@ -516,32 +516,25 @@ The following procedure can be used to verify a subtree consistency proof.
...
-   3. Until `LSB(fn)` is set or `sn` is `0`, right-shift `fn` and `sn` equally.
-   4. For each value `p` in the `proof` array:
-      1. If `sn` is `0`, then stop iteration and fail the proof verification.
-      2. Set `r` to `HASH(0x01 || p || r)`.
-      3. Until `LSB(sn)` is set, right-shift `sn`.
-      4. Right-shift `sn` once more.
-   5. Compare `sn` to `0` and `r` to `root_hash`. If either is not equal, fail the proof verification. If all are equal, accept the proof.
-2. Otherwise, run the following:
+1. Set `fn` to `start`, `sn` to `end - 1`, and `tn` to `n - 1`.
+1. If `sn` is `tn`, then:
```
lukevalenta, line 520: We could collapse lines 520-523:
```suggestion
1. Right-shift `fn`, `sn`, and `tn` equally until either `fn` is `sn` or both `sn` is not `tn` and `LSB(sn)` is not set.
```

lukevalenta: And maybe swap the until to the start for consistency with below (or switch below to be consistent with this). Although maybe this doesn't read as well:
```suggestion
1. Until either `fn` is `sn` or both `sn` is not `tn` and `LSB(sn)` is not set, right-shift `fn`, `sn`, and `tn` equally.
```

lukevalenta (approved):
Nice refactor! Just a couple comments.

draft-davidben-tls-merkle-tree-certs.md
lukevalenta: I think the ‘both’ helps to make it unambiguous (and Gemini agreed), but it does take a little more brain power to process so I’m fine with keeping it separated out.

Ooh yeah that’s a good point about checking if sn is tn initially vs in the loop condition.

PR #150, opened 2025-10-06
Add an appendix that explains the proof procedures

### LINKED ISSUE #147: Explain the proof verification procedures in an appendix
We borrowed them from 9162, which also didn't explain them, but it's probably worth doing so.

davidben:
CC @nharper
    
Closes #147

### DIFF
--- a/draft-davidben-tls-merkle-tree-certs.md
+++ b/draft-davidben-tls-merkle-tree-certs.md
@@ -1728,7 +1728,7 @@ Inclusion proofs can also be evaluated by considering these two stages separatel
 
 A subtree consistency proof for `[start, end)` and the tree of `n` elements is similar to an inclusion proof for element `end - 1`. If one starts from `end - 1`'s hash, incorporating the whole inclusion proof should reconstruct `root_hash` and incorporating a subset of the inclusion proof should reconstruct `node_hash`. Thus `end - 1`'s hash and this inclusion proof can prove consistency. A subtree consistency proof in this document applies two optimizations over this construction:
 
-1. Instead of starting at level 0 with `end - 1`, the proof can start at a higher level. Any ancestor of `end - 1` shared by both the subtree and the overall tree will `node_hash` and `root_hash`. Use the highest level with a commmon ancestor. This truncates the inclusion proof portion of the consistency proof.
+1. Instead of starting at level 0 with `end - 1`, the proof can start at a higher level. Any ancestor of `end - 1` shared by both the subtree and the overall tree is a valid starting node to reconstruct `node_hash` and `root_hash`. Use the highest level with a commmon ancestor. This truncates the inclusion proof portion of the consistency proof.
 
 2. If this starting node is the entire subtree, omit its hash from the consistency proof. The verifier is assumed to already know `node_hash`.
 


draft-davidben-tls-merkle-tree-certs.md
```diff
@@ -1423,6 +1629,165 @@ at-trustAnchorID ATTRIBUTE ::= {
...
+
+## Consistency Proof Verification {#consistency-proof-bits}
+
+The procedure in {{verifying-a-subtree-consistency-proof}} iteratively builds two hashes, `fr` and `sr`, which are expected to equal `node_hash` and `root_hash`, respectively. Everything hashed into `fr` is also hashed into `sr`, so success demonstrates that `root_hash` contains `node_hash`.
+
+A subtree consistency proof for `[start, end)` and the tree of `n` elements is very similar to an inclusion proof for element `end - 1`. Starting at this element, incorporating the whole inclusion proof should reconstruct `root_hash` and incorporating a subset of the inclusion proof should reconstruct `node_hash`. However, instead of starting at level 0, the proof can be optimized to start at a higher level, specifically the highest ancestor of `end - 1` that is also directly contained in the tree.
```
nharper: There are some ambiguities in this. The first is describing the "very similar to an inclusion proof": it's not clear that "starting at this element" means "include element `end - 1` in the proof" (from all of the examples I've looked at of consistency proofs that go down to level 0, they include two sibling elements, whereas an inclusion proof doesn't).

The second is "specifically the highest ancestor of `end - 1`". Ancestor isn't defined in this document. Perhaps add a definition of ancestor (at the beginning of the doc), e.g. a valid subtree `[start, end)` is an ancestor of leaf node `i` if `start <= i < end`.

The third is "that is also directly contained in the tree". I assume that "the tree" is "the tree of `n` elements" rather than the unnamed subtree "`[start, end)`". With that phrase being the only qualifier on the highest ancestor, that highest ancestor becomes the root of the tree of `n` elements, when it should be an interior node that is common to both trees.

Possible rewording:

A subtree consistency proof for the subtree `[start, end)` and the tree of `n` elements is very similar to an inclusion proof for element `end - 1`. A proof can be constructed by adding the element `end - 1` to the inclusion proof of `end - 1` in the tree. This new proof allows reconstructing `root_hash`, and a subset of that proof can be used to reconstruct `node_hash`. Depending on the structure of the subtree, instead of starting this proof at level 0, the proof can be optimized to start a higher level. The consistency proof starts at the highest level of an ancestor of `end - 1` shared by both the subtree and the tree.

draft-davidben-tls-merkle-tree-certs.md
```diff
@@ -1423,6 +1629,165 @@ at-trustAnchorID ATTRIBUTE ::= {
...
+
+Steps 3 and 4 then skip to the starting node, described above. The starting node may be:
+
+* The entire subtree `[start, end)` if `[start, end)` is directly contained in the tree. This will occur if `end` is `n`, or if `[start, end)` is full.
+
+* Otherwise, the highest full subtree along the right edge of `[start, end)`.
+
+In the first case, `fn` will equal `sn` after truncation. Step 5 will then initialize the hashes to `node_hash`. The consistency proof does not need to include a separate copy of `node_hash`.
+
+In the second case, `fn` is less than `sn`. Step 6 will then initialize the hashes to the first value in the consistency proof.
```
nharper: It's not clear what cases these are referring to, and they're not the two bullet points above. (`fn` can equal `sn` after performing either of those steps.) I'd clarify by changing this to something like the following:

> Steps 5 and 6 initialize the hashes `fr` and `sr`.
>
> If `fn` equals `sn` (possibly after truncation), step 5 initializes the hashes to `node_hash` because the subtree `[start, end)` is directly contained in the tree, so the consistency proof did not need to include a separate copy of `node_hash`.
>
> Otherwise, `fn` is less than `sn`, and step 6 initializes the hashes to the first value in the consistency proof.

draft-davidben-tls-merkle-tree-certs.md
```diff
@@ -1423,6 +1629,165 @@ at-trustAnchorID ATTRIBUTE ::= {
...
+|[0,2)| |[2,4)| |[4,6)|   level 1
++-----+ +-----+ +-----+
+  / \     / \     / \
++-+ +-+ +-+ +-+ +-+ +-+
+|0| |1| |2| |3| |4| |5|   level 0
++-+ +-+ +-+ +-+ +-+ +-+
+~~~
+{: #fig-merkle-tree-bits-partial title="An example Merkle Tree of size 6"}
+
+When the size of a Merkle Tree is not a power of two, some levels on the rightmost edge of the tree are skipped. The rightmost edge is the path to the last element. The skipped levels can be seen in its binary representation. Here, the last element is 5, which has binary representation `0b101`. When a bit is set, the corresponding node is a right child. When it is unset, the corresponding node is skipped.
```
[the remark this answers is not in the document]
davidben: Hmm. Right subtree makes it sound like this is possible:

```
                +----------------------------+
                |            [0, 11)         |   level 4
                +----------------------------+
                   /                  |
       +----------------+             |
       |     [0, 8)     |             *          level 3
       +----------------+             |
        /              \              |
   +--------+      +--------+   +---------+
   | [0, 4) |      | [4, 8) |   | [8, 11) |      level 2
   +--------+      +--------+   +---------+
    /      \        /      \     |      |
+-----+ +-----+ +-----+ +-----+  |   +------+
|[0,2)| |[2,4)| |[4,6)| |[6,8)|  *   |[8,11)|    level 1
+-----+ +-----+ +-----+ +-----+  |   +------+
  / \     / \     / \     / \    |    /  \
+-+ +-+ +-+ +-+ +-+ +-+ +-+ +-+ +-+ +-+ +--+
|0| |1| |2| |3| |4| |5| |6| |7| |8| |9| |10|     level 0
+-+ +-+ +-+ +-+ +-+ +-+ +-+ +-+ +-+ +-+ +--+
```

Was trying to capture that it was specifically the right edge of the patch that skips things. It's true that they're also skipped for other *paths*, but only if they haven't diverged from the right edge before then. The later paragraph "Zero bits also indicate..." covers that.

draft-davidben-tls-merkle-tree-certs.md
```diff
@@ -1423,6 +1629,165 @@ at-trustAnchorID ATTRIBUTE ::= {
...
+|[0,2)| |[2,4)| |[4,6)| |     |  level 1
++-----+ +-----+ +-----+ +-----+
+  / \     / \     / \     / \
++-+ +-+ +-+ +-+ +-+ +-+ +-+ +-+
+|0| |1| |2| |3| |4| |5| | | | |  level 0
++-+ +-+ +-+ +-+ +-+ +-+ +-+ +-+
+~~~
+{: #fig-merkle-tree-bits-partial-comparison title="An example Merkle Tree of size 6, viewed as a subset of a tree of size 8"}
+
+Zero bits also indicate skipped nodes in paths that have not yet diverged from the rightmost edge. In the example, the binary representation of 4 is `0b100`. While bit 0 and bit 1 are both unset, they manifest in the tree differently. Bit 0 indicates that 4 is a right child. However, at bit 1, `0b100` has not yet diverged from the last element, `0b101`. That instead indicates a skipped node, not a left child.
```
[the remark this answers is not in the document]
davidben: Hehe, that's true, I definitely generally think of paths are going from root to leaf in trees. Though, you're right, you do more naturally construct it leaf to root. I'm not sure which direction readers would find intuitive by default.

I switched to direction-neutral terminology in hopes that readers can just mentally fill in the direction they like better. :-)

draft-davidben-tls-merkle-tree-certs.md
nharper: Your wordsmithing looks good.

nharper (approved):
one nit, otherwise looks good

PR #158, opened 2025-11-04
Additional motivation: reduce the cost of log operation

dkg:
this was motivated by discussion in the PLANTS BoF today.  PQ signing algorithms are not the only additional pressure on log operators.


davidben:
Works for me! Not sure what the process looks like post-BoF but hopefully that'll become clear.

PR #159, opened 2025-11-05
charter: elaborate that private PKIs might not need transparency

bwesterb:
Does this work for you @johngray-dev @ounsworth as an alternative to removing transparency earlier on?


debcooley:
I like this update....

PR #161, opened 2025-11-06
Restore the more efficient trust anchor negotiation

davidben:
When trust_anchors was extracted from the first MTC drafts, we lost an optimization because it was not, at the time, relevant for the baseline version of the problem. However, it is still quite useful for MTCs (and may be useful for other scenarios too).

This implements https://github.com/tlswg/tls-trust-anchor-ids/issues/62 to put it back. For now, it's in the MTC repo, but in the long run, we should put it back in TAI itself.

### DIFF
--- a/draft-davidben-tls-merkle-tree-certs.md
+++ b/draft-davidben-tls-merkle-tree-certs.md
@@ -1408,7 +1408,12 @@ The following procedure can be used to perform this check. It succeeds if `r` co
 * The certification path's trust anchor ID appears in the relying party's `trust_anchors` extension, or
 * One of the certification path's additional trust anchor ranges contains some ID in the relying party's `trust_anchors` extension
 
-In applications that use additional trust anchor ranges, relying parties MAY send a single trust anchor ID to represent all certificates whose trust anchor ranges contain that trust anchor ID.
+Trust anchor ranges do not impact an authenticating party's list of available trust anchors in EncryptedExtensions (see {{Section 4.3 of !I-D.ietf-tls-trust-anchor-ids}}) or the HTTPS/SVCB record (see {{Section 5 of !I-D.ietf-tls-trust-anchor-ids}}). Those continue to reference the single trust anchor ID that corresponds to each certificate.
+
+In applications that use additional trust anchor ranges, relying parties MAY send a single trust anchor ID to represent all certificates whose trust anchor ranges contain that trust anchor ID. This includes:
+
+* Trust anchors that are sent in response to an EncryptedExtensions or HTTPS/SVCB message from the authenticating party
+* Trust anchors that are sent in `trust_anchors`, independently of the authenticating party
 
 ## Using Trust Anchor IDs
 


matt-mueller:
Should there be something said about how the DNS negotiation & the retry mechanism work with ranges?

Since the DNS hint could be stale, the client would probably want to match the prefix against its known base_ids, but send the base_id.max for highest chance of success without requiring a round trip. For the retry mechanism the client could do a specific match, but for simplicity it would be nice to just have the client do the same thing in both cases.

draft-davidben-tls-merkle-tree-certs.md
```diff
@@ -1352,19 +1352,72 @@ When a CA is found to be untrustworthy, relying parties SHOULD remove trust in t
...
+* What the relying party should communicate to the authenticating party to help it make this decision
+
+Certificate selection in TLS, described in {{Section 4.4.2.2 and Section 4.4.2.3 of !RFC8446}}, incorporates both explicit relying-party-provided information in the ClientHello and CertificateRequest messages and implicit deployment-specific assumptions. This section describes a RECOMMENDED integration of Merkle Tree certificates into TLS trust anchor IDs ({{!I-D.ietf-tls-trust-anchor-ids}}), but applications MAY use application-specific criteria in addition to, or instead of, this recommendation.
+
+## Extensions to Trust Anchor IDs
+
+[[TODO: Move this into draft-ietf-tls-trust-anchor-ids once the PLANTS WG is further along. See https://github.com/tlswg/tls-trust-anchor-ids/issues/62]]
+
+A TLS deployment may know that all relying parties that accept one trust anchor must additionally accept another trust anchor, or desire identifiers for groups of related trust anchors. For example, in this document, any relying party which recognizes some landmark is known to additionally accept earlier landmarks, up to the `max_landmarks` parameter.
```
bwesterb: > any relying party which recognizes some landmark is known to additionally accept earlier landmarks, up to the `max_landmarks` parameter.

Nit: by induction a relying will then accept all previous landmarks if *up to* isn't read as a bound.

What about:

> For example, in this document, the relying party will recognize up to `max_landmark` consecutive landmarks.

draft-davidben-tls-merkle-tree-certs.md
```diff
@@ -1352,19 +1352,72 @@ When a CA is found to be untrustworthy, relying parties SHOULD remove trust in t
...
+struct {
+    TrustAnchorID base;
+    uint64 min;
+    uint64 max;
+} TrustAnchorRange;
+
+TrustAnchorRange TrustAnchorRangeList<1..2^16-1>;
+~~~
+
+A trust anchor range `r` is said to *contain* a trust anchor ID `id`, if `id`, as a relative OID, is the concatenation of `r.base` and some integer component between `min` and `max`. The following procedure succeeds if `r` contains `id` and fails otherwise:
```
bwesterb: Add "inclusive".

draft-davidben-tls-merkle-tree-certs.md
```diff
@@ -1352,19 +1352,72 @@ When a CA is found to be untrustworthy, relying parties SHOULD remove trust in t
...
 
-[[TODO: We can do slightly better. If the relying party supports landmark 42, it can be assumed to also support landmark 41, 40, 39, etc. https://github.com/tlswg/tls-trust-anchor-ids/issues/62 discusses how to fit this into the trust anchor IDs framework. This allows the client to summarize its state with one ID per CA.]]
+A relying party that has been configured with trusted subtrees ({{trusted-subtrees}}) derived from a set of landmarks SHOULD configure the `trust_anchors` extension to advertise the highest supported landmark in the set. The selection procedures defined in {{!I-D.ietf-tls-trust-anchor-ids}} and {{!extensions-to-trust-anchor-ids}} will then correctly determine whether a signatureless certificate is compatible with the relying party.
```
bwesterb: It will be common for servers to configure two signatureless certs. If I read it correctly, we're not telling how to pick between them. It seems sensible to sent the signatureless cert with the most recent landmark, as an older one might be expired.

bwesterb (approved):

draft-davidben-tls-merkle-tree-certs.md
cjpatton: Maybe add a sentence here explaining what the procedure is doing.

davidben: Done

davidben: Reworded slightly.

draft-davidben-tls-merkle-tree-certs.md
```diff
@@ -1352,19 +1352,72 @@ When a CA is found to be untrustworthy, relying parties SHOULD remove trust in t
...
+    uint64 min;
+    uint64 max;
+} TrustAnchorRange;
+
+TrustAnchorRange TrustAnchorRangeList<1..2^16-1>;
+~~~
+
+A trust anchor range `r` is said to *contain* a trust anchor ID `id`, if `id`, as a relative OID, is the concatenation of `r.base` and some integer component between `min` and `max`. The following procedure succeeds if `r` contains `id` and fails otherwise:
+
+1. Check that the most-significant bit of the last byte of `r.base` is unset. If it is set, fail the procedure.
```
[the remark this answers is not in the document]
davidben: Done

davidben (commented):
> Should there be something said about how the DNS negotiation & the retry mechanism work with ranges?

Done

> Since the DNS hint could be stale, the client would probably want to match the prefix against its known base_ids, but send the base_id.max for highest chance of success without requiring a round trip. For the retry mechanism the client could do a specific match, but for simplicity it would be nice to just have the client do the same thing in both cases.

Agreed. Tried to clarify this in the new text.

PR #167, opened 2025-12-12
Address part of Éric Vyncke's comments

bwesterb:
https://mailarchive.ietf.org/arch/msg/plants/h56f-ERetlQ2Qr_9q5V2FxIL_JI/


davidben:
Ah yeah, that's awkward. How about https://github.com/davidben/merkle-tree-certs/pull/168?

PR #174, opened 2026-01-30
Various minor fixes

aww-aww:


draft-davidben-tls-merkle-tree-certs.md
```diff
@@ -186,31 +181,29 @@ informative:
...
-This document describes Merkle Tree certificates, a new form of X.509 certificates which integrate public logging of the certificate, in the style of Certificate Transparency. The integrated design reduces logging overhead in the face of both shorter-lived certificates and large post-quantum signature algorithms, while still achieving comparable security properties to traditional X.509 and Certificate Transparency. Merkle Tree certificates additionally admit an optional signatureless optimization, which decreases the message size by avoiding signatures altogether, at the cost of only applying to up-to-date relying parties and older certificates.
+This document introduces Merkle Tree certificates, a new form of X.509 certificate that integrates public logging directly into the issuance process, similar to Certificate Transparency. This design reduces the overhead caused by short-lived certificates and large post-quantum signature algorithms while maintaining traditional security standards. Additionally, it offers an optional 'signatureless' mode that further decreases message size by removing signatures entirely for older certificates when validated by up-to-date relying parties.
```
davidben: > Additionally, it offers an optional 'signatureless' mode that further decreases message size by removing signatures entirely for older certificates when validated by up-to-date relying parties.

I think this changes the meaning of the sentence subtly in the wrong way. The base mechanism decreases the size of the *logs* but not the certificates[*]. (Well, cosignature negotiation might give some levers we can pull here, but broadly no.) The optional optimization decreases the size of the certificates.

[*] Certificate used loosely. In CT with TLS sometimes the SCTs are sent next to the certificate, rather than directly in the certificate. I suspect that's why I said "message size", but that is a bit ambiguous.

davidben: > similar to Certificate Transparency

This could be read to mean that the integration is similar to CT, when it's really the public logging is similar to CT. Admittedly the old phrasing isn't great here either, but the new phrasing I think goes further in the ambiguous direction.

draft-davidben-tls-merkle-tree-certs.md
```diff
@@ -186,31 +181,29 @@ informative:
...
 
-Additionally, as PKIs transition to shorter-lived certificates {{CABF-153}} {{CABF-SC081}}, the number of entries in the log will grow.
-
-This document introduces Merkle Tree certificates, a new form of X.509 certificate that integrates logging with certificate issuance. Each CA maintains a log of everything it issues, signing views of the log to assert it has issued the contents. The CA signature is combined with cosignatures from other parties who verify correct operation and optionally mirror the log. These signatures, together with an inclusion proof for an individual entry, constitute a certificate.
+Instead of attaching separate SCTs, a CA maintains a signed log of everything it issues. The certificate consists of a log entry, a CA signature over the log view, and an inclusion proof. The CA signature is combined with cosignatures from other parties who verify correct operation and optionally mirror the log. These signatures, together with an inclusion proof for an individual entry, constitute a certificate.
```
davidben: This loses the glue text, which seems important here? It's unclear when we've switched from talking about CT to talking about MTCs.

draft-davidben-tls-merkle-tree-certs.md
```diff
@@ -338,31 +331,31 @@ In Certificate Transparency, a CA first certifies information by signing it, the
...
+3. Logging: The CA operates an append-only *issuance log* ({{issuance-logs}}). Unlike a CT log, this issuance log only contains entries added by the CA:
 
    1. The CA adds a TBSCertificateLogEntry ({{log-entries}}) to its log, describing the information it is certifying.
 
-   2. The CA signs a *checkpoint*, which describes the current state of the log. A signed checkpoint certifies that the CA issued *every* entry in the Merkle Tree ({{certification-authority-cosigners}}).
+   2. The CA signs a *checkpoint* to certify that the CA issued *every* entry in the Merkle Tree to that point ({{certification-authority-cosigners}}).
 
-   3. The CA additionally signs *subtrees* ({{subtrees}}) that together contain certificates added since the last checkpoint ({{arbitrary-intervals}}). This is an optimization to reduce inclusion proof sizes. A signed subtree certifies that the CA has issued *every* entry in the subtree.
+   3. The CA may also sign *subtrees* ({{subtrees}}) containing all certificates added since the last checkpoint ({{arbitrary-intervals}}). This optimization reduces inclusion proof sizes. A signed subtree certifies that the CA has issued *every* entry in the subtree.
```
davidben: I think we can s/may also sign/signs/

Yes, you can technically pick other subtrees or even the root and everything will still work. But the specification is currently prescriptive that this is what you do. (Possibly we should change that, but that's probably a different discussion.)

draft-davidben-tls-merkle-tree-certs.md
```diff
@@ -338,31 +331,31 @@ In Certificate Transparency, a CA first certifies information by signing it, the
...
 
-A certificate with cosignatures is known as a *full certificate*. Analogous to X.509 trust anchors and trusted CT logs, relying parties are configured with trusted cosigners ({{trusted-cosigners}}) that allow them to accept Merkle Tree certificates. The inclusion proof proves the TBSCertificate is part of some subtree, and cosignatures from trusted cosigners prove the subtree was certified by the CA and available to monitors. Where CT logs entire certificates, the issuance log's entries are smaller TBSCertificateLogEntry ({{log-entries}}) structures, which do not scale with public key or signature size.
+A certificate with cosignatures is a *full certificate*. Analogous to X.509 trust anchors and trusted CT logs, relying parties configure trusted cosigners ({{trusted-cosigners}}) that allow them to accept Merkle Tree certificates. Inclusion proofs prove TBSCertificates are part of subtree, and cosignatures from trusted cosigners prove the subtree was certified by the CA and available to monitors. CT logs entire certificates, whereas the issuance log's entries are smaller TBSCertificateLogEntry ({{log-entries}}) structures. These do not scale with public key or signature size.
```
davidben: "relying parties configure trusted cosigners" is a bit ambiguous whether RPs have a list of trusted cosigners (correct) or RPs change the config of trusted cosigners (incorrect)

The latest couple sentence now reads a little roundabout to me now because it's missing the couple of glue words to say that the first clause is is comparing to CT, not expanding on the above. Agreed that the original text is also a bit odd, but I'm not sure the new text is in the right direction.

draft-davidben-tls-merkle-tree-certs.md
```diff
@@ -939,6 +932,8 @@ When `type` is `tbs_cert_entry`, `N` is the number of bytes needed to consume th
 
 `tbs_cert_entry_data` contains the contents octets (i.e. excluding the initial identifier and length octets) of the DER {{X.690}} encoding of a TBSCertificateLogEntry, defined below. Equivalently, `tbs_cert_entry_data` contains the DER encodings of each field of the TBSCertificateLogEntry, concatenated. This construction allows a single-pass implementation in {{verifying-certificate-signatures}}.
 
+The `tbs_cert_entry_data` field contains the Value portion of the DER-encoded TBSCertificateLogEntry. It omits the initial Identifier (Tag) and Length octets that usually wrap a DER sequence. It is created by concatenating the individual DER encodings of every field within the TBSCertificateLogEntry in order. This construction allows {{verifying-certificate-signatures}} in a single pass, increasing efficiency.
```
davidben: Did you mean to replace the above paragraph? It looks like you've duplicated it.

That said, I don't believe this replacement text is correct. Take a look at X.680 and X.690. X.690 does not call that portion the "Value". It calls it the "contents". A "value" in ASN.1 is the abstract data value, not the contents octets. From X.680:

> A data type (or type for short) is a category of information (for example, numeric, textual, still image or video information). A data value (or value for short) is an instance of such a type

Since many folks work with ASN.1 without having read the formal spec and knowing the formal terminology, the original text has a parenthetical "(i.e excluding ...)" and another helper "Equivalently, ...". But I think the first definition should be the one that is precise: the contents octets of the DER-encoded TBSCertificateLogEntry.

Also {{verifying-certificate-signatures}} will expand to "Section ???". The new text doesn't work grammatically. It will say "allows Section ??? in a single pass" reads very strangely. The original text was written with the expansion in mind.

davidben (commented):
Thanks! I haven't gotten all the way through this PR yet, but flushing comments now. Some of the changes make sense, while others seem to make the text slightly wrong, which makes it a bit slow to go through.

draft-davidben-tls-merkle-tree-certs.md
```diff
@@ -995,7 +990,7 @@ struct {
...
+A cosigner always tracks the checkpoint with the largest end index, designating it as their current checkpoint. This represents the maximum extent of the log the cosigner has verified. To ensure consistency, a cosigner is restricted in what subtrees they can sign. A cosigner MUST NOT sign any subtree [start, end) if its end point exceeds the tree_size of their current checkpoint (unless start is 0). A cosigner may only sign a "middle" or "end" portion of a tree if that portion is already contained within the current verified checkpoint. Every signature a cosigner produces must be provably consistent with their current checkpoint via a subtree consistency proof. By signing a subtree, the cosigner becomes responsible for all entries in the tree up to that end point—even if the full checkpoint is missing or unavailable.
```
bwesterb: Change in meaning: cosigner doesn't know what's latest end index—it just retains the one.

bwesterb:
Hi Andy, thanks for this. Majority of your _many_ edits are great and make things easier to read. But with quite a few you also change the meaning. We could pull those out if you like.

aww-aww:
Please yank anything that does not make sense. I tried to guess when things
were ambiguous to me, but it looks a number of the guesses went the wrong
direction.

bwesterb:
> Please yank anything that does not make sense. I tried to guess when things were ambiguous to me, but it looks a number of the guesses went the wrong direction.

Hi @aww-aww, we merged many of your suggestions from this PR in #188, #189, #190 and #191.

davidben:
@aww-aww anything else from this PR you think we should merge, or should we close this PR out?

aww-aww:
The useful bits have been pulled. Closing

PR #181, opened 2026-02-08
Initial changes based on Michael Richardson's review

bwesterb:


draft-davidben-tls-merkle-tree-certs.md
```diff
@@ -344,11 +344,11 @@ Merkle Tree Certificates are issued as follows. {{fig-issuance-overview}} depict
 
 3. The CA operates an append-only *issuance log* ({{issuance-logs}}). Unlike a CT log, this issuance log only contains entries added by the CA:
 
-   1. The CA adds a TBSCertificateLogEntry ({{log-entries}}) to its log, describing the information it is certifying.
+   a. The CA adds a TBSCertificateLogEntry ({{log-entries}}, abbreviated "tbscert entries" in the diagram) to its log, describing the information it is certifying.
```
davidben: Markdown nitpick: This ends up making paragraphs that happen to start with "a.", "b.", etc. We want an ordered list that is styled with letters instead of numbers. I believe the syntax is to keep it with `1.` etc and instead stick `{: type="a"}` right before the list, like this:

```
   {: type="a"}
   1. blah blah blah.
   2. blah blah blah.
   3. blah blah blah.
```

(`{: whatever}` is some kramdown-rfc syntax for setting random XML attributes and then the [rfcxml ol element](https://authors.ietf.org/en/rfcxml-vocabulary) has a type attribute.)

draft-davidben-tls-merkle-tree-certs.md
```diff
@@ -974,7 +974,7 @@ A cosigner computes a cosignature for a subtree in some log by signing a MTCSubt
 ~~~tls-presentation
 opaque HashValue[HASH_SIZE];
 
-/* From Section 4.1 of draft-ietf-tls-trust-anchor-ids */
+/* From Section 3.1 of draft-ietf-tls-trust-anchor-ids */
```
davidben: I think this was correct as-is?
https://www.ietf.org/archive/id/draft-ietf-tls-trust-anchor-ids-02.html#section-4.1

(The one below seems to indeed be section 8 though.)

PR #187, opened 2026-02-24
Rename to standalone and landmark certificates

### LINKED ISSUE #163: Consider better names for full and signatureless certificate

davidben:
Closes #163

### DIFF
--- a/draft-ietf-plants-merkle-tree-certs.md
+++ b/draft-ietf-plants-merkle-tree-certs.md
@@ -186,7 +186,7 @@ informative:
 
 --- abstract
 
-This document describes Merkle Tree certificates, a new form of X.509 certificates which integrate public logging of the certificate, in the style of Certificate Transparency. The integrated design reduces logging overhead in the face of both shorter-lived certificates and large post-quantum signature algorithms, while still achieving comparable security properties to traditional X.509 and Certificate Transparency. Merkle Tree certificates additionally admit an optional signatureless optimization, which decreases the message size by avoiding signatures altogether, at the cost of only applying to up-to-date relying parties and older certificates.
+This document describes Merkle Tree certificates, a new form of X.509 certificates which integrate public logging of the certificate, in the style of Certificate Transparency. The integrated design reduces logging overhead in the face of both shorter-lived certificates and large post-quantum signature algorithms, while still achieving comparable security properties to traditional X.509 and Certificate Transparency. Merkle Tree certificates additionally admit an optional size optimization that avoids signatures altogether, at the cost of only applying to up-to-date relying parties and older certificates.
 
 --- middle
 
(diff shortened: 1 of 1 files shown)


bwesterb (approved):

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1434,27 +1434,27 @@ In applications that use additional trust anchor ranges, relying parties MAY sen
...
 
-The full certificate response SHOULD additionally carry a alternate URL for the signatureless certificate, as described {{Section 7.4.2 of !RFC8555}}. Before the signatureless certificate is available, the alternate URL SHOULD return a HTTP 503 (Service Unavailable) response, with a Retry-After header ({{Section 10.2.3 of !RFC9110}}) estimating when the certificate will become available. Once the next landmark is allocated, the ACME server constructs a signatureless certificate, as described in {{signatureless-certificates}} and serves it from the alternate URL.
+The standalone certificate response SHOULD additionally carry a alternate URL for the landmark certificate, as described {{Section 7.4.2 of !RFC8555}}. Before the landmark certificate is available, the alternate URL SHOULD return a HTTP 503 (Service Unavailable) response, with a Retry-After header ({{Section 10.2.3 of !RFC9110}}) estimating when the certificate will become available. Once the next landmark is allocated, the ACME server constructs a landmark certificate, as described in {{landmark-certificates}} and serves it from the alternate URL.
```
```suggestion
The standalone certificate response SHOULD additionally carry an alternate URL for the landmark certificate, as described {{Section 7.4.2 of !RFC8555}}. Before the landmark certificate is available, the alternate URL SHOULD return a HTTP 503 (Service Unavailable) response, with a Retry-After header ({{Section 10.2.3 of !RFC9110}}) estimating when the certificate will become available. Once the next landmark is allocated, the ACME server constructs a landmark certificate, as described in {{landmark-certificates}} and serves it from the alternate URL.
```

lukevalenta (approved):
I really like this edit. It makes the relationship between landmark certificates and landmark tree sizes very clear.

initsecret (approved):
reads well to me.

on a tangent, the terminology section got me wondering if a standalone cert would ever present an inclusion proof to a _not_-landmark subtree. That seems analogous to signing an end certificate with the root certificate.

davidben:
> on a tangent, the terminology section got me wondering if a standalone cert would ever present an inclusion proof to a not-landmark subtree. That seems analogous to signing an end certificate with the root certificate.

I expect inclusion proof to non-landmark subtree to be the default case for standalone certs. :-) Part of the point of a standalone cert is to that you can get a cert immediately (imagine having to rekey in a hurry). That means you don't want to wait for a landmark.

The way I've imagined this is that you have two checkpoint feeds, a fast one and a slow (landmark) one. Your fast one runs about as fast as you're able to generate signatures and ping witnesses/mirrors. Let's say that's every 30 seconds. So every 30 seconds, you mint a new fast checkpoint, and collect cosigs. Then you can generate standalone certificates based on the subtree(s) between the last fast checkpoint and the current one. If you look at it funny, this is kind of like a batch signing scheme.

Then there's a slow feed of checkpoints which is every couple of hours (the landmarks). Now, you *could* use that feed to generate standalone certs too, but there's not much point. You may as well use the fast feed. The only reason for the slow feed is so the total number of live checkpoints is small and predistributable. So we use those to generate the landmark certificates, without cosignatures.

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -2141,3 +2141,5 @@ In draft-04, there is no fast issuance mode. In draft-05, frequent, non-landmark
 {:numbered="false"}
 
 - Address editorial comments from WG adoption call
+
+- Renamed full certificate to standalone certificate, signatureless certificate to landmark certificate.
```
aww-aww, line 2145: nit: s/,/and/

PR #193, opened 2026-03-06
Rename 'max_landmarks' to 'max_active_landmarks' for clarity, and fix example

lukevalenta:
* Rename `max_landmarks` to `max_active_landmarks` for clarity, since CAs also need to keep one 'inactive' landmark around in the landmark sequence, and that could lead to implementation confusion.
* Fix example of setting max landmarks. The formula in the recommended procedure for allocating landmarks is `max_landmarks = ceil(max_cert_lifetime / time_between_landmarks) + 1`. The `+ 1` accounts for landmarks not allocated at the exact start of their time interval, which can push certificate expiry one interval further than `ceil(max_cert_lifetime / time_between_landmarks)` alone would bound.

  For example, say we have:

      time_between_landmarks: 1, giving time intervals [0,1), [1,2), [2,3)
      max_cert_lifetime: 3
      landmarks at times: 0.99, 1, 2, 3

  The landmark at time 0.99 could contain certs that don't expire until time 3.99. Thus, we need to keep all 4 landmarks active.

### DIFF
--- a/draft-ietf-plants-merkle-tree-certs.md
+++ b/draft-ietf-plants-merkle-tree-certs.md
@@ -1214,7 +1214,7 @@ A landmark certificate is constructed based on a *landmark sequence*, which is a
 A landmark sequence has the following fixed parameters:
 
 * `base_id`: An OID arc for trust anchor IDs of individual landmarks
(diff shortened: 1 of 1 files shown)


davidben (approved):
Ah yeah that's a better name.

PR #200, opened 2026-03-19
Relax restrictions on null_entry

davidben:
While mostly pointless, adding random null_entry types in the middle of a log isn't harmful, per se. We restricted because we didn't have a reason, but it might be useful if:

- A CA wishes to do a periodic health check on any external cosigners it works with.

- A monitoring ecosystem wants the CA to append to a log with *at least* some frequency.

Realistically, the CA probably has enough issuance rate to make both of these moot but, eh, we may as well have one less restriction, and then people don't have to think about this.


bwesterb (approved):

davidben:
@FiloSottile and @lukevalenta point out that I misunderstood tlog, and that both logs and witnesses can bump the timestamp on their signatures when they feel like it. Less clearly motivated, but probably an unnecessary restriction too. 🤷

lukevalenta (approved):

PR #203, opened 2026-04-10
Another spelling of the ACME integration

### LINKED ISSUE #202: HTTP 503 isn't quite right for pending landmark
Watson Ladd points out:

> I'm not sure 503 is the right error code for the landmark-relative cert in ACME: a 400 series code would make more sense given that some are about resource conditions. There isn't one that fits though. 503 may complicate infrastructure monitoring as it is usually a server problem you want to worry about.

https://mailarchive.ietf.org/arch/msg/plants/-Q5kuTg9HQUTcbkv8ybIndgqj5U/

davidben:
Instead of messing with the URL, this other idea proposes to update the ACME order object, using a very similar flow as the ACME "processing" state.

Fixes #202

@aarongable WDYT?


draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1463,15 +1463,39 @@ When both a landmark and standalone certificate are supported by a relying party
...
+
+When processing an order for a Merkle Tree certificate, the ACME server moves the order to the "valid" state once the corresponding entry is sequenced in the issuance log. The order's certificate URL then serves the standalone certificate, constructed as described in {{standalone-certificates}}. If the ACME server supports landmark-relative certificates, it SHOULD return an updatable order ({{updatable-acme-orders}}) and set the Retry-After header field to an estimate of when the landmark-relative certificate will be available.
+
+When the landmark-relative certificate is available, the ACME server updates the order with a new certificate URL. The new certificate URL SHOULD continue to serve the standalone certificate, but it SHOULD additionally carry an alternate URL ({{Section 7.4.2 of !RFC8555}}) for the landmark-relative certificate. The alternate URL SHOULD include trust anchor ID information as above.
```
bwesterb: This can also be used to reanchor the standalone cert to the landmark. Do we want to mention it?

bwesterb (approved):

draft-ietf-plants-merkle-tree-certs.md
lukevalenta: Right, I think we want ACME clients to be able to serve three certs:

1) The original standalone cert, for its short inclusion proof.
2) The landmark-relative cert, for the landmark optimization.
3) The landmark-anchored standalone cert (option (2) but with signatures), for the server storage optimization.

The ACME server only needs to make (1) and (3) available, since option (2) can be derived from option (3) at TLS handshake time by stripping the signatures.

bwesterb: Deriving (2) from (3) requires some intelligence on the ACME client side.

davidben: Hehehe, this did occur to me but probably how/whether to mention reanchoring can be a separate change.

I'm not actually sure ACME should reanchor, at least not automatically. The reanchored certificate is bigger. It's mostly a trade-off for really big deployments that don't want to store both certs. I guess it could be a separate profile that the ACME server offers? Or it could be something that big deployments assemble themselves.

lukevalenta: Makes sense. I suppose "really big deployments" can just get the landmark subtrees the same way that RPs do.

lukevalenta (approved):

draft-ietf-plants-merkle-tree-certs.md
aarongable: Offering the reanchored cert feels like a perfect use of the link rel=alternate mechanism that already exists. They're both available at the same time, but only some clients will want it.

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1463,15 +1463,39 @@ When both a landmark and standalone certificate are supported by a relying party
...
 # ACME Extensions
 
-This section describes how to issue Merkle Tree certificates using ACME {{!RFC8555}}.
+## Updatable ACME Orders
 
-When downloading the certificate ({{Section 7.4.2 of !RFC8555}}), ACME clients supporting Merkle Tree certificates SHOULD send "application/pem-certificate-chain-with-properties" in their Accept header ({{Section 12.5.1 of !RFC9110}}). ACME servers issuing Merkle Tree certificates SHOULD then respond with that content type and include trust anchor ID information as described in {{Section 6 of !I-D.ietf-tls-trust-anchor-ids}}. {{use-in-tls}} decribes the trust anchor ID assignments for standalone and landmark-relative certificates.
+A successful ACME {{!RFC8555}} order ends in the "valid" state, at which point the certificates are available for download and the order object is completed. However, in cases such as landmark-relative Merkle Tree certificate, the order may be in a partially complete state where both:
+
+* Some complete set of certificates are available for download, and the ACME client can begin deploying certificates
+* Additional optional certificates may be available later
```
aarongable: editorial:
```suggestion
* some complete set of certificates are available for download, and the ACME client can begin deploying certificates; and
* additional optional certificates may be available later.
```

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1463,15 +1463,39 @@ When both a landmark and standalone certificate are supported by a relying party
...
 
-When processing an order for a Merkle Tree certificate, the ACME server moves the order to the "valid" state once the corresponding entry is sequenced in the issuance log. The order's certificate URL then serves the standalone certificate, constructed as described in {{standalone-certificates}}.
+ACME clients and servers used with Merkle Tree certificates SHOULD support:
 
-The standalone certificate response SHOULD additionally carry an alternate URL for the landmark-relative certificate, as described {{Section 7.4.2 of !RFC8555}}. Before the landmark-relative certificate is available, the alternate URL SHOULD return a HTTP 503 (Service Unavailable) response, with a Retry-After header ({{Section 10.2.3 of !RFC9110}}) estimating when the certificate will become available. Once the next landmark is allocated, the ACME server constructs a landmark-relative certificate, as described in {{landmark-relative-certificates}} and serves it from the alternate URL.
+* Certificate properties ({{Section 6 of !I-D.ietf-tls-trust-anchor-ids}})
+* Updatable orders ({{updatable-acme-orders}})
+* Alternate certificates chains ({{Section 7.4.2 of !RFC8555}})
```
```suggestion
* certificate properties ({{Section 6 of !I-D.ietf-tls-trust-anchor-ids}});
* updatable orders ({{updatable-acme-orders}}); and
* alternate certificates chains ({{Section 7.4.2 of !RFC8555}}).
```

draft-ietf-plants-merkle-tree-certs.md
aarongable: There are two possible axes of change when an Order is updated:
- Does the certificate URL change?
- Do the contents of the certificate URL change?

I think this PR has selected the worst quadrant of that 2x2 grid.

If the URL doesn't change and the contents don't change, that's just ACME today.

If the URL doesn't change but the contents _do_ change, that's a little confusing (since URLs should generally point to static resources, and presumably the old standalone cert would have to be moved to a different alternate URL) but at least it would match the behavior established by ACME STAR.

If the URL changes but the contents don't change, as you've proposed here... what happens to that original URL? Should the CA still honor it? Why are we allocating two different unique identifiers for the same cert? Is the URL change just to indicate to the client that a diff has actually happened? If so, that should instead be conveyed via a `lastUpdated` field or something like that.

If the URL changes and the contents also change, that makes the most sense to me. We have a new certificate, here's the new URL and contents, and if you want the old one again it's still accessible via link rel=alternate. I don't have concerns about clients getting confused -- if it is sufficiently updated to poll an `updateable` order, it can also be smart enough to follow the alternate links.

davidben: The contents change in that the response headers are different, and that's part of the HTTP response, cached, etc. Though you're right the response body hasn't changed.

We *could* decide the landmark cert is the first one and the (presumably already downloaded) standalone cert is the alternate, which the client could then recognize, but then we risk another problem:

Suppose an alternates-naive client requests a cert. And it happens that a landmark was just about to be allocated. The client takes long enough to that actually as soon as it first sees the order, both certs are available. CA puts standalone second per this... but now the client *only* deploys the landmark cert and has a problem.

davidben: Ah sorry I replied in a hurry (very hectic day) and missed some of what you wrote:

> If the URL changes but the contents don't change, as you've proposed here... what happens to that original URL? Should the CA still honor it?

Probably. At least briefly because there might be a race condition between the two requests.

(Mostly I assumed the extra URL wasn't too expensive but I could be wrong. I mean it could be as simple as a cache-busting query parameter on the same URL.)

> Why are we allocating two different unique identifiers for the same cert? Is the URL change just to indicate to the client that a diff has actually happened? If so, that should instead be conveyed via a lastUpdated field or something like that.

Ah sure. URL + lastUpdated works for me too. I do think it's worth preserving the invariant that the first URL is the more default (standalone) certificate, and that we *somehow* signal a change to the client, but that works too! I'll update (hah) that later today.

davidben: > Ah sure. URL + lastUpdated works for me too. I do think it's worth preserving the invariant that the first URL is the more default (standalone) certificate, and that we somehow signal a change to the client, but that works too! I'll update (hah) that later today.

Done. How's this?

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1463,15 +1463,44 @@ When both a landmark and standalone certificate are supported by a relying party
...
+
+When an order has a usable set of certificates available but may be updated in the future, the ACME server SHOULD set the order to the "valid" state, with the `updatable` field set to true. HTTP responses for the order SHOULD include a Retry-After header field ({{Section 10.2.3 of !RFC9110}}) to suggest a polling interval to the client. After the final update to the order, the server SHOULD set the `updatable` field to false, indicating that the order is fully complete. When updating an order or certificate, an ACME server implementing this extension MUST set the order's `lastUpdated` field to the time of the update.
+
+A client that receives an updatable order SHOULD send an ACME POST-as-GET request ({{Section 6.3 of !RFC8555}}) after the time given in the Retry-After header field ({{Section 10.2.3 of !RFC9110}}) of the response, if any. If the updated order contains a new `lastUpdated` value, the client SHOULD repeat downloading the certificates ({{Section 7.4.2 of !RFC8555}}) and replace the previously-downloaded certificates. The client SHOULD repeat this process until the `updatable` field is no longer true.
```
ilaril: If update adds landmark-relative certificate, then no certificate should be replaced. Trying to replace existing standalone certificate with landmark-relative one is highly likely to cause errors, and even replacing existing landmark-relative one is likely to degrade performance. What the client should do in this case is adding a new landmark-relative certificate to the line (and cleaning up any expired ones).

The only situations that I can come up with where replacing existing certificates would make sense is if more preferred alternative (landmark-relative certificates are **not** alternates) appears, or if certificate lifespan is extended (ACME STAR).

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1463,15 +1463,44 @@ When both a landmark and standalone certificate are supported by a relying party
...
 
-The standalone certificate response SHOULD additionally carry an alternate URL for the landmark-relative certificate, as described {{Section 7.4.2 of !RFC8555}}. Before the landmark-relative certificate is available, the alternate URL SHOULD return a HTTP 503 (Service Unavailable) response, with a Retry-After header ({{Section 10.2.3 of !RFC9110}}) estimating when the certificate will become available. Once the next landmark is allocated, the ACME server constructs a landmark-relative certificate, as described in {{landmark-relative-certificates}} and serves it from the alternate URL.
+When the landmark-relative certificate is available, the ACME server updates the order to serve new certificates. The order's certificate URL SHOULD continue to serve the standalone certificate, but it SHOULD now additionally carry an alternate URL ({{Section 7.4.2 of !RFC8555}}) for the landmark-relative certificate. The alternate URL SHOULD include trust anchor ID information as above.
```
ilaril: "alternate" is not correct relation for this case (one can not use landmark-relative certificate instead of standalone one, one gets errors if one tries that).

Additionally, RFC 8555 says: "certificate resources are immutable once issuance is complete". ACME STAR does not change that, it defines its own resource.

draft-ietf-plants-merkle-tree-certs.md
davidben: The updated order is still expected to serve the standalone certificate, and is still expected to serve the standalone certificate first (see the section immediately following this one). The last paragraph explains why this is important.

draft-ietf-plants-merkle-tree-certs.md
davidben: > "alternate" is not correct relation for this case (one can not use landmark-relative certificate instead of standalone one, one gets errors if one tries that).

Alternate is the relation used in RFC 8555. See https://www.rfc-editor.org/rfc/rfc8555.html#section-7.4.2

RFC 8555 specifically uses the example of using alternates to describe paths to different trust anchors. That sounds like an exactly analogous situation. The alternates are not, a priori, substitutable with each other because they work in different relying parties. Likewise, the difference between a landmark-relative and standalone certificate is that they work in different relying parties.

Perhaps RFC 8555 should not have used alternate as the relation, but I think that ship has sailed. 🤷 

> Additionally, RFC 8555 says: "certificate resources are immutable once issuance is complete". ACME STAR does not change that, it defines its own resource.

Ah good call. The original version of this PR said you updated the URL, but I switched it to this after @aarongable's comments above. Given that, it sounds like we should go back to the previous version and ask that ACME servers mint new URLs when they update the set.

draft-ietf-plants-merkle-tree-certs.md
davidben: (Running around a bit this morning but later today I'll update this to also make the general section clear that you're expected to always serve a complete set of certificates.)

draft-ietf-plants-merkle-tree-certs.md
ilaril: Well, for RFC 8555, the intention is that the ACME client chooses one, so in sense these are alternates. That is not the case with standalone/landmark-relative certificates.

Additionally, web linking RFC recommends that relations with highly application-specific meaning should be named accordingly. I think this relation has very application-specific meaning.

draft-ietf-plants-merkle-tree-certs.md
ilaril: Serving a complete set might not be an option, as the old landmark-relative certificate can still be valid, but have different keys. That certificate should be retained until it expires.

davidben: Right, the old landmark-relative certificate came from a different order and is not in scope for replacement here. This text is talking about replacing the certificates from *this* order.

It's also true that optimal server behavior is to retain the previous order until it expires, but that's external to an individual ACME order.

davidben: > (Running around a bit this morning but later today I'll update this to also make the general section clear that you're expected to always serve a complete set of certificates.)

Done

draft-ietf-plants-merkle-tree-certs.md
davidben: > Ah good call. The original version of this PR said you updated the URL, but I switched it to this after @aarongable's comments above. Given that, it sounds like we should go back to the previous version and ask that ACME servers mint new URLs when they update the set.

I've updated and switched this back.

> Well, for RFC 8555, the intention is that the ACME client chooses one, so in sense these are alternates. That is not the case with standalone/landmark-relative certificates.

Hmm. The text in RFC 8555 says:

> The server MAY provide one or more link relation header fields
>    [[RFC8288](https://www.rfc-editor.org/rfc/rfc8288)] with relation "alternate".  Each such field SHOULD express
>   an alternative certificate chain starting with the same end-entity
>   certificate.  This can be used to express paths to various trust
>   anchors.  Clients can fetch these alternates and use their own
>   heuristics to decide which is optimal.

I read this to mean that the client fetches both alternates and then uses "heuristics" to decide which to use in any given TLS connection. (And then TAI aims to fix these heuristics into something well-defined.)

If I understand you right, you read this to mean that the client fetches both alternates and uses "heuristics" to decide which to install and which to throw away, never to be used at all.

I agree that, by your read, this feature is inappropriate to use for any kind of multi-certificate story and we'd need something else. However, it seems that to me that the second version makes for a pretty useless feature! I was not involved when ACME was being standardized, so I cannot say what was the "intention". Perhaps @aarongable would have some insight.

> Additionally, web linking RFC recommends that relations with highly application-specific meaning should be named accordingly. I think this relation has very application-specific meaning.

I do not think this needs to be that application-specific, or rather not MTC-specific. The general primitive we have here and in quite a lot of PKI scenarios is:

* ACME server gives you multiple certificate chains that together cover the certificate request
* You might get multiple certificate chains to meet relying party variability, you might get them because some chains are smaller + less generally useful (omit cross-signs, landmark-relative, depend on a smaller signature algorithm, etc.)
* ACME server gives you enough information, in some well-defined format, that you can toss it into your TLS server and dispatch the right cert to the right RP (CertificatePropertyList)

Whether that's `alternate` or a new relation depends on the question above. If ACME defined `alternate` to be the first interpretation, this all seems perfectly fine. In that outcome, perhaps ACME should have named it something else, but I don't think it'd be worth renaming now. If ACME defined `alternate` to be the second interpretation then, yeah, we should pick another word. Meh.

draft-ietf-plants-merkle-tree-certs.md
davidben: Forgot to mention, updated version also makes it clear that replacement is only within the scope of this one order.

In particular, I think you misunderstood what I meant by "complete set". When I said "complete set" I meant "a set of certificates that satisfies this client's order. The client asked for "give me newly-issued certificates, for this key, that work with RPs that trust this CA". Serving just a landmark certificate, as you suggested in the first comment would not be a complete set and was never what this PR was suggesting. Valid responses to the order are standalone or standalone+landmark-relative. The old landmark-relative certificate has nothing to do with *this* order and as you note may not have the same key, so they should not be served here. No one was suggesting that they be served here.

The AP may well want to retain the old landmark-relative certificate from the previous order. That is outside of the scope of one ACME order. The reason we need a complete set for *this order* is explained later in the PR, discussing a race condition. From there, the reason we need replacement is that, if you are serving standalone+landmark, you don't want the AP to end up with two copies of the _same_ standalone certificate.

This mechanism also allows things like the ACME server updating the CertificatePropertyList for a certificate once it has more information on bulk-ID inclusions, etc. That is another case where we need replacement.

(Think this thread is resolved, but I'll wait for your reply in case there's still a disconnect.)

draft-ietf-plants-merkle-tree-certs.md
ilaril: > I read this to mean that the client fetches both alternates and then uses "heuristics" to decide which to use in any given TLS connection. (And then TAI aims to fix these heuristics into something well-defined.)

(ACME) clients are (usually :-) ) not TLS servers.

> If I understand you right, you read this to mean that the client fetches both alternates and uses "heuristics" to decide which to install and which to throw away, never to be used at all.

Yes. Besides the text, installing multiple requires specialized TLS server support and has edge cases where it is not clear what is sensible behavior.
 
> I agree that, by your read, this feature is inappropriate to use for any kind of multi-certificate story and we'd need something else. However, it seems that to me that the second version makes for a pretty useless feature!

It is useful for choosing the root certificate, and Let's Encrypt has used (and I think is currently using) it for that purpose.

> > Additionally, web linking RFC recommends that relations with highly application-specific meaning should be named accordingly. I think this relation has very application-specific meaning.
> 
> I do not think this needs to be that application-specific, or rather not MTC-specific. 

ACME clients need to handle MTC landmark-relative certificates differently from stand-alone certificates (MTC or X.509). For example, replacing standalone certificates but not landmark-relative ones.

>     * ACME server gives you enough information, in some well-defined format, that you can toss it into your TLS server and dispatch the right cert to the right RP (CertificatePropertyList)

If order can give multiple used certificates and certificates can be replaced, there is also the question of which exact certificate should be replaced.

draft-ietf-plants-merkle-tree-certs.md
ilaril: > This mechanism also allows things like the ACME server updating the CertificatePropertyList for a certificate once it has more information on bulk-ID inclusions, etc. That is another case where we need replacement.

Bulk-ID inclusions that change seem to be prone to cause failures (AP thinks bulk-ID means something the RP does not). And even immutable bulk IDs can require AP to have additional information.

davidben: > Bulk-ID inclusions that change seem to be prone to cause failures (AP thinks bulk-ID means something the RP does not).

Not what I'm suggesting. There's goofy ordering thing with bulk IDs across CAs and landmarks, which metadata update can solve. But this is a digression this point and not necessarily a lever we want to actually pull anyway. (We can debate whether to allow for optional bulk IDs at all in the TAI discussion.)

> And even immutable bulk IDs can require AP to have additional information.

Which is why TAI defines CertificatePropertyList to carry this information.

It sounds like there are no more issues on this thread that are directly related to this PR. I'll go ahead and mark this resolved.

draft-ietf-plants-merkle-tree-certs.md
davidben: Will leave the question of whether to use alternate or make a new relation type for parallel certificates to ACME experts. I care about it being suitably generic but I don't care what the spelling is.

Though it's also orthogonal to this PR so we can resolve that separately if it takes too long to resolve. I can file a bug for that so we don't forget, once I'm at a real keyboard. :-) This PR was mostly about dealing with the HTTP 503 situation.

> ACME clients need to handle MTC landmark-relative certificates differently from stand-alone certificates (MTC or X.509). For example, replacing standalone certificates but not landmark-relative ones.

MTC certificates are X.509 certificates. :-p

The cross-order replacement thing is a difference but they do not *need* to do that. Nothing *breaks* if they don't. It *is* better if they keep the old one around, but they don't have to differentiate the two if they don't want. Keeping the previous order's standalone certificate is unnecessary, but isn't *harmful* because it'll just be shadowed by the new one. Or...

* One could just parse the certificate and notice it's a landmark-relative one
* We could define a "I think you should keep this around" property if we want it to be generic
* The ACME client can notice that the standalone certificates overlap in TAI metadata while the landmark-relative ones do not

(If we need a new ACME mechanism we can do another PR after fixing the HTTP 503 issue, but my inclination is that "don't bother differentiating" or "parse the certificate if you care" are pretty good.)

> If order can give multiple used certificates and certificates can be replaced, there is also the question of which exact certificate should be replaced.

The PR answers this, no? Within the bounds of the order update story, the replacement is for the certificates that came from this order. That seems natural to me.

The PR says nothing about cross-order replacement, which I think is correct for stuff relating to one order. See above for whether we need a mechanism or not.

davidben: Thinking about it, I'm not sure the two interpretations of "alternate" are *necessarily* incompatible. ACME already says you have to use a heuristic to decide which chain to use. If you implemented it with a heuristic, in a scheme where you have to commit to only one chain to install, your heuristic presumably should say "don't use the landmark certificate" because you're not able to.

If you *are* able to install both and evaluate the heuristic by ClientHello, the same mechanism is just as serviceable for a smarter setup.

And if you're able to pick up CertificatePropertyList and TAI for a well-defined selection, you don't need to maintain a heuristic.

That all feels like different degrees on how to use what is really the same mechanism: CA has a couple way to prove what you asked for and you have to evaluate the size vs applicability trade-off.

But again I don't have a horse in this race. If it's the consensus of ACME folks that there should be two of these, that's cool too.

ilaril: MTCs are not X.509 certificates. The format might be the same, but the semantics are very different.

Unfortunately, neither CPL nor TAI replaces heuristics due to compatibility constraints. And those heuristics need to be configured to TLS server (and editing random config files is no-no). Well, MTC might be easy: Flag everything as ignored by heuristics.

davidben: Let's leave discussion of the negotiation bits to those threads. With all this jumping around between threads and vague references to things that you don't think work, jumping to a different topic each time, it is too hard to follow.

I really do want to understand and address your concerns. But to do that, we have to be able to dive into each concern individually and figure out where you misunderstand the design (and the text could be clearer) and where we've missed something important. That's not possible with all this jumping around.

This PR is about adjusting how ACME models the funny valid-but-processing-like state in the issuance flow. *What* it delivers and plugs into server behavior seems pretty orthogonal. This one PR is not meant to be the one PR that finishes the spec in one go.

davidben: (As promised, I've filed https://github.com/ietf-plants-wg/merkle-tree-certs/issues/206 for the link relation.)

ilaril: Why would ACME client replace existing certificate on update (other than due relative certificate slot limits)?

davidben: The first download contained new-standalone-cert. The second download contains (new-standalone-cert, new-landmark-relative-cert). You should replace the first download rather than having two copies of new-standalone-cert.

The second download needs to have both because of this:

> In normal operation, an ACME client will first observe the order only serving a standalone certificate, and later the updated order serving both certificates. However, if the certificate is issued shortly before a landmark is allocated, the landmark-relative certificate will be available very soon after the standalone certificate. Depending on timing, an ACME client might then only observe the updated order. Such a client depends on the updated order serving a complete set certificates, not just the newly-added ones.

Why _wouldn't_ the ACME client replace the existing certificates from _this order_? It seems natural to me that, if the mechanism has the CA update the result of _this order_, the correct semantics are that the previous certificates from _this order_ are correspondingly replaced on the client. (Keep in mind the previous landmark-relative certificate is from a _previous_ order and wouldn't be replaced.)

ilaril: On the second download, the ACME client already has stand-alone certificate from that order. It sees the landmark-relative one as new, so it adds it (unless slot limits force replacement).

davidben: Ah sure. I think we're using different words to describe the same outcome?

We can't use the same URL (at least as long as we use link headers) because, as you noted, the URLs are immutable. So even changing response headers to add a link header counts as a change.

That means the second copy of the standalone certificate needs to be deduped against the first copy somehow. We could say "look for an identical certificate and replace it", but then we need to answer stupid questions like "how much do you need to compare". ISTM it's cleaner to just say it's a replacement of the old set with the new set. Yes, the old set is a subset of the new set, so we're effectively adding. But saying replace means the ACME client doesn't have to compare the contents of the sets. It also gives us room to play games like:

- Here's a landmark-relative cert, it's part of these group IDs
- Oh I heard back from this root program, it's also part of this group ID

The replacement model then nicely handles this. (Whether we want to do that, I don't know. This is just a convenient bonus that falls out of what seems to be the simpler model anyway?)

jsha:
Thanks for proposing this. I've written up a third way to do it: https://github.com/ietf-plants-wg/merkle-tree-certs/pull/207. Please let me know what you think!

davidben:
Closing since this is now irrelevant with the other one.

PR #204, opened 2026-04-10
Clarify issuer field in TBSCertificateLogEntry

CBonnell:
- Removed `issuer` from "as-is" list of fields
- Fixed formatting of `issuer` description. Also changed PKIX to X.500 to align with RFC 5280.

### DIFF
--- a/draft-ietf-plants-merkle-tree-certs.md
+++ b/draft-ietf-plants-merkle-tree-certs.md
@@ -959,11 +959,13 @@ TBSCertificateLogEntry ::= SEQUENCE {
 
 The fields of a TBSCertificateLogEntry are defined as follows:
 
-* `version`, `issuer`, `validity`, `subject`, `issuerUniqueID`, `subjectUniqueID`, and `extensions` have the same semantics as the corresponding TBSCertificate fields, defined in {{Section 4.1.2 of !RFC5280}}.
+* `version`, `validity`, `subject`, `issuerUniqueID`, `subjectUniqueID`, and `extensions` have the same semantics as the corresponding TBSCertificate fields, defined in {{Section 4.1.2 of !RFC5280}}.
+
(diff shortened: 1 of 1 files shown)


CBonnell:
I can add X.500 as an informative reference, or can do that as another PR. Thoughts?

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -959,11 +959,13 @@ TBSCertificateLogEntry ::= SEQUENCE {
 
 The fields of a TBSCertificateLogEntry are defined as follows:
 
-* `version`, `issuer`, `validity`, `subject`, `issuerUniqueID`, `subjectUniqueID`, and `extensions` have the same semantics as the corresponding TBSCertificate fields, defined in {{Section 4.1.2 of !RFC5280}}.
+* `version`, `validity`, `subject`, `issuerUniqueID`, `subjectUniqueID`, and `extensions` have the same semantics as the corresponding TBSCertificate fields, defined in {{Section 4.1.2 of !RFC5280}}.
+
+*  `issuer` is the issuance log's log ID as a X.500 distinguished name, as described in {{log-ids}}.
```
davidben: [There was feedback](https://mailarchive.ietf.org/arch/msg/plants/rQPkX4tEQySSvMzsH1bciLkSL8o/) on the list to call this a "PKIX distinguished name" instead of "X.50[09] distinguished name". I have no horse in this race, but I guess we like to say PKIX instead? I've no idea.

davidben (approved):
Thanks!

draft-ietf-plants-merkle-tree-certs.md
CBonnell: Thanks for the pointer. Yeah, I don't know why that suggestion was made, because the DN as defined in 5280 is the X.500 DN. Nonetheless, this is a minor detail we can all discuss later, so I'll change it back to PKIX for now.

PR #207, opened 2026-04-13
acme: poll landmarkURL, POST newRelativeCertificate

jsha:
An ACME client polls landmarkURL (which may be shared across many certificates) to determine when a landmark-relative certificate is possible. Once a landmark-relative certificate is possible, the ACME client POSTs to newRelativeCertificate (from the directory).

Advantages:

 - No new states in the Order state machine.
 - No new polling semantics for certificate download URLs.
 - CAs do no extra work for clients that don't want relative certificates (or: clients can defer getting relative certificates until they see that traffic justifies it).
 - Order lifetime does not depend on landmarking period.
 - All HTTP requests are expected to serve 200 in the happy case.
 - The new field in the directory serves as a signal that the ACME server supports Merkle Tree Certificates.

Disadvantages:

 - ACME clients will tend to rush in and request relative certificates as soon as a landmark is minted. Options:
   - CA returns 503 if load is too high and clients backoff and retry.
   - Some mechanism for clients to decide how long to wait after a landmark is minted before requesting a relative certificate. Perhaps interpolation based on size of landmarks?

Claude drafted the IANA Considerations section, which I reviewed and reference checked. All other words are mine.

### DIFF
--- a/draft-ietf-plants-merkle-tree-certs.md
+++ b/draft-ietf-plants-merkle-tree-certs.md
@@ -1465,13 +1465,48 @@ When both a landmark and standalone certificate are supported by a relying party
 
 This section describes how to issue Merkle Tree certificates using ACME {{!RFC8555}}.
 
-When downloading the certificate ({{Section 7.4.2 of !RFC8555}}), ACME clients supporting Merkle Tree certificates SHOULD send "application/pem-certificate-chain-with-properties" in their Accept header ({{Section 12.5.1 of !RFC9110}}). ACME servers issuing Merkle Tree certificates SHOULD then respond with that content type and include trust anchor ID information as described in {{Section 6 of !I-D.ietf-tls-trust-anchor-ids}}. {{use-in-tls}} decribes the trust anchor ID assignments for standalone and landmark-relative certificates.
+Order objects reach the "valid" state once a standalone certificate is issued. The standalone certificate is made available at the "certificate" URL in the order object. No additional states for order objects are defined.
 
-When processing an order for a Merkle Tree certificate, the ACME server moves the order to the "valid" state once the corresponding entry is sequenced in the issuance log. The order's certificate URL then serves the standalone certificate, constructed as described in {{standalone-certificates}}.
+Two new fields are defined:
 
-The standalone certificate response SHOULD additionally carry an alternate URL for the landmark-relative certificate, as described {{Section 7.4.2 of !RFC8555}}. Before the landmark-relative certificate is available, the alternate URL SHOULD return a HTTP 503 (Service Unavailable) response, with a Retry-After header ({{Section 10.2.3 of !RFC9110}}) estimating when the certificate will become available. Once the next landmark is allocated, the ACME server constructs a landmark-relative certificate, as described in {{landmark-relative-certificates}} and serves it from the alternate URL.
+ - In the directory, a "newRelativeCertificate" field.
+ - In order objects, a "landmarkURL" field.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1465,13 +1465,48 @@ When both a landmark and standalone certificate are supported by a relying party
...
 
-When processing an order for a Merkle Tree certificate, the ACME server moves the order to the "valid" state once the corresponding entry is sequenced in the issuance log. The order's certificate URL then serves the standalone certificate, constructed as described in {{standalone-certificates}}.
+Two new fields are defined:
 
-The standalone certificate response SHOULD additionally carry an alternate URL for the landmark-relative certificate, as described {{Section 7.4.2 of !RFC8555}}. Before the landmark-relative certificate is available, the alternate URL SHOULD return a HTTP 503 (Service Unavailable) response, with a Retry-After header ({{Section 10.2.3 of !RFC9110}}) estimating when the certificate will become available. Once the next landmark is allocated, the ACME server constructs a landmark-relative certificate, as described in {{landmark-relative-certificates}} and serves it from the alternate URL.
+ - In the directory, a "newRelativeCertificate" field.
```
lukevalenta, line 1472: I'd go with the full name here.
```suggestion
 - In the directory, a "newLandmarkRelativeCertificate" field.
```

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1465,13 +1465,48 @@ When both a landmark and standalone certificate are supported by a relying party
...
+ - In order objects, a "landmarkURL" field.
```
lukevalenta, line 1473: If the client is `landmarkURL`-aware, another direction we could go is to make clients fully [tlog-tiles](c2sp.org/tlog-tiles)-aware and have them reconstruct their landmark-relative inclusion proofs by directly fetching the needed tiles from the CA issuance log (or its mirrors). So instead of `landmarkURL` we'd pass `issuanceLogURL` (`<prefix>` from tlog-tiles), and maybe some of its mirrors.

Pros:
1) No ACME server API is needed for processing landmark-relative certs and serving them to clients. This cuts the ACME server out of the landmark-relative certificate reconstruction process entirely, and removes the need for the `newRelativeCertificate` field in the directory. Now we can just use vanilla RFC 8555 plus the directory field.
2) Clients fetch what they need themselves from the static, cacheable tiles, thus solving the "thundering herd" problem mentioned in the PR description. Everything can be served from a CDN.
3) ACME clients that want landmark-anchored standalone certs would need to be tile-aware anyway.

Cons (that I can think of):
1) ACME clients now need to be aware of the tlog-tiles serving infrastructure, and we were trying to keep these things decoupled. We could perhaps solve this by saying that ACME is _only_ concerned with standalone certificates, and for everything else the client needs to interact with the CA issuance log.
2) Fetching individual tiles to reconstruct an inclusion proof to a landmark subtree will be a little more complex on the ACME client side (but there are existing tlog-tiles libraries that could be reused), and would require issuing multiple requests and downloading some extraneous data.
3) For ecosystems without transparency, the CA issuance log might not serve tiles publicly so you may want to fetch things from an authenticated CA API. This is also a problem for this PR's proposal, since `landmarkURL` probably points to the CA issuance log's S3-compatible storage bucket anyway (at least in our implementation, it does). This isn't a problem for the WebPKI, at least.
4) Clients that want landmark-anchored standalone certs will need to find a way to fetch the necessary CA and cosigner signatures on the landmark subtrees separately (the same as the other proposals).

davidben:
Guessing we should close this now?

PR #209, opened 2026-04-14
Clarify that CRL and OCSP apply unchanged to Merkle Tree certificates

### LINKED ISSUE #208: Explicitly state that CRL and OCSP are unmodified
MTC doesn't touch OCSP or CRL. If you were using CRLs before, you can continue to use CRLs as before. If you weren't using CRLs before, you can continue to not use CRLs as before.

We should say this explicitly so it's clear. (There might be interesting things to do in this space, but I think we can say this is orthogonal.)

sgmenda:
Closes #208

### DIFF
--- a/draft-ietf-plants-merkle-tree-certs.md
+++ b/draft-ietf-plants-merkle-tree-certs.md
@@ -1386,7 +1386,7 @@ A misbehaving CA might correctly construct a globally consistent log, but refuse
 
 When a CA is found to be untrustworthy, relying parties SHOULD remove trust in that CA. To minimize the compatibility impact of this mitigation, index-based revocation can be used to only distrust entries after some index, while leaving existing entries accepted. This is analogous to the {{SCTNotAfter}} mechanism used in some PKIs.
 
-Index-based revocation is complementary to existing certificate-level revocation mechanisms like CRLs {{!RFC5280}} and OCSP {{!RFC6960}}, which apply unchanged to Merkle Tree certificates.
+The revocation mechanism in this section is complementary to certificate-level revocation mechanisms. Because Merkle Tree certificates use log indices as serial numbers, existing revocation mechanisms like CRLs {{!RFC5280}} and OCSP {{!RFC6960}} apply unchanged.
 
 # Use in TLS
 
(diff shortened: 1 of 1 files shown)


draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1386,6 +1386,8 @@ A misbehaving CA might correctly construct a globally consistent log, but refuse
 
 When a CA is found to be untrustworthy, relying parties SHOULD remove trust in that CA. To minimize the compatibility impact of this mitigation, index-based revocation can be used to only distrust entries after some index, while leaving existing entries accepted. This is analogous to the {{SCTNotAfter}} mechanism used in some PKIs.
 
+Index-based revocation is complementary to existing certificate-level revocation mechanisms like CRLs {{!RFC5280}} and OCSP {{!RFC6960}}, which apply unchanged to Merkle Tree certificates.
```
davidben: Confusingly, those actually are, themselves, index-based revocation because we have cleverly stuck the indices into the serial number. Though, hmm, it's true that we phrased it as "the relying party maintains a list of revoked ranges of indices", which is a different animal from CRLs and OCSP.

Not sure how best to capture that subtlety in here.

davidben (commented):
Thanks!

I tried to find whether we said anything like "don't use a cosigner key with another protocol" but neither I nor the word salad machine could find it, so I think that's okay by the text. In terms of domain separation, our label doesn't start with SEQUENCE (0x20) so we're good there too.

draft-ietf-plants-merkle-tree-certs.md
davidben: ("This text is fine, no need to complicate it" is also a fine answer.)

sgmenda: ah, astute. how about:

```suggestion
The revocation mechanism in this section is complementary to certificate-level revocation mechanisms. Because Merkle Tree certificates use log indices as serial numbers, existing revocation mechanisms like CRLs {{!RFC5280}} and OCSP {{!RFC6960}} apply unchanged.
```

davidben: LGTM!

davidben (approved):

sgmenda:
@davidben On cosigner key separation, the closest I can find in the spec is this line: https://github.com/ietf-plants-wg/merkle-tree-certs/blob/586c9e13daa7f6e2b63ec81ea366d13a20c0a3db/draft-ietf-plants-merkle-tree-certs.md?plain=1#L1048

Agree that we should be explicit, would you like me to open a PR for it?

davidben:
Ah yeah, that might be worth clarifying. PR sounds good! Thanks!

davidben:
BTW, I've pushed https://github.com/ietf-plants-wg/merkle-tree-certs/commit/6b34c98dfd33493ead59a004bf9b7e0fade519f7 to list you and this contribution.

PR #210, opened 2026-04-15
Recommend against reusing cosigner keys for other purposes

sgmenda:
See #208 discussion.

### DIFF
--- a/draft-ietf-plants-merkle-tree-certs.md
+++ b/draft-ietf-plants-merkle-tree-certs.md
@@ -111,6 +111,14 @@ informative:
     - name: Henry Faltin Våge
     - org: University of Bergen
 
+  KeyReuse:
+    title: "Security in the Presence of Key Reuse: Context-Separable Interfaces and their Applications"
+    target: https://eprint.iacr.org/2019/519
+    date: 2019
+    author:
+    - name: Christopher Patton
+    - name: Thomas Shrimpton
+
   STH-Discipline:
     title: STH Discipline & Security Considerations
     target: https://mailarchive.ietf.org/arch/msg/trans/Zm4NqyRc7LDsOtV56EchBIT9r4c/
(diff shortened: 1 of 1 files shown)


draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -111,6 +111,14 @@ informative:
...
     - org: University of Bergen
 
+  KeyReuse:
+    title: "Security in the Presence of Key Reuse: Context-Separable Interfaces and their Applications"
+    target: https://eprint.iacr.org/2019/519
+    date: 2019
+    author:
+    - name: Christopher Patton
+    - name: Thomas Shrimpton
+
```
sgmenda, line 121: I tried using a DOI, but rendered bib entry was too noisy:

> Patton, C. and T. Shrimpton, "Security in the Presence of Key Reuse: Context-Separable Interfaces and Their Applications", Springer International Publishing, Lecture Notes in Computer Science pp. 738-768, DOI 10.1007/978-3-030-26948-7_26, ISBN ["9783030269470", "9783030269487"], 2019, <https://doi.org/10.1007/978-3-030-26948-7_26>. 

So I switched to this instead.

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -979,6 +987,8 @@ A cosignature MAY implicitly make additional statements about a subtree, determi
 
 Each cosigner has a public key and a *cosigner ID*, which uniquely identifies the cosigner. The cosigner ID is a trust anchor ID {{!I-D.ietf-tls-trust-anchor-ids}}. By identifying the cosigner, the cosigner ID specifies both the public key and the additional statements made by the cosigner's signatures. If a single operator performs multiple cosigner roles in an ecosystem, each role MUST use a distinct cosigner ID and SHOULD use a distinct key.
 
+Following the principle of key separation (see, e.g., {{KeyReuse}}), cosigner keys SHOULD NOT be used for purposes outside this document.
```
davidben, line 990: One can probably read this to ban OCSP and CRLs. We assert in security considerations that they apply unchanged, but I'm not sure if we've officially said they're allowed. Maybe "Certification Authority Cosigners" should do that.

I do wonder if we should add something just to leave open the door for similar allowed reuses where they've been judged OK? In the old MTC, before the massive tlog/X.509 rewrite, we wrote:

> Extensions of this protocol MAY be defined which reuse the keys, but any that do MUST use a different label string and analyze the security of the two uses concurrently.

"MUST use a different label string" is probably not right because OCSP and CRLs don't have a label string, but perhaps we can say something like the other uses MUST NOT overlap with the `label` prefix in {{signature-format}} or something? I dunno.

(Would "Extensions of this protocol and specific cosigner roles" make more sense here?)

davidben (commented):
Thanks!

draft-ietf-plants-merkle-tree-certs.md
sgmenda: Good point, lemme think on it. 🤔

davidben: WDYT about https://github.com/ietf-plants-wg/merkle-tree-certs/pull/224?

PR #211, opened 2026-04-17
subtree: define Full, Partial, and Right Subtree

jsha:
The previous language said a subtree "is" an interior node. It seemed clearer to say that it is an interior node plus all of its children.

I wound up defining three named kinds of subtrees, which I think are helpful in conceptualizing what's going on: Full Subtree, Right Subtree, and Partial Subtree. Then we can say early on that we'll be using a Full Subtree combined with a Right Subtree to cover arbitrary ranges.

Also, I defined a Right Subtree `[start, n)` to only be contained in the Merkle Tree of size `n`. In the tree of size `n+1`, at least one of the nodes has a different hash, so the Right Subtree `[start, n)` isn't really contained in it. This allows us to remove some language around "directly contained," which I found confusing. Note: I haven't yet updated the consistency proof section, since I wanted to first see if folks found this tweak compelling.

### DIFF
--- a/draft-ietf-plants-merkle-tree-certs.md
+++ b/draft-ietf-plants-merkle-tree-certs.md
@@ -296,6 +296,12 @@ Standalone certificate:
 Landmark-relative certificate:
 : An optimized certificate containing an inclusion proof to a landmark subtree, and no signatures.
 
+Successor:
+: A Merkle Tree of size N is the successor of all Merkle Trees of size < N with which it is consistent.
+
+Predecessor:
+: A Merkle Tree of size N is the predecessor of all Merkle Trees of size > N with which it is consistent.
+
 # Overview
 
 In Certificate Transparency, a CA first certifies information by signing it, then submits the resulting certificate (or precertificate) to logs for logging. Merkle Tree Certificates invert this process: the CA certifies information by logging it, then submits the log to cosigners to verify log operation. A certificate is assembled from the result and proves the information is in the CA's log.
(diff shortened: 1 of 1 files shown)


draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -407,30 +413,44 @@ Landmark-relative certificates are constructed and used as follows. {{fig-landma
...
+## Right Subtrees
+
+A Right Subtree contains all the children of one of the nodes on the right-hand edge of a Merkle Tree of a specific size.  In a Merkle Tree of size `n`, all subtrees identified by `[A, n)` are Right Subtrees, and all Right subtrees are identified by `[A, n)` for some `A`.
+
+## Partial Subtrees
+
+Partial Subtrees only occur along the right-hand edge of a Merkle Tree. In other words, all Partial Subtrees are also Right Subtrees.
 
-The *size* of the subtree is `end - start`. If the subtree's size is a power of two, it is said to be *full*, otherwise it is said to be *partial*.
+In a Merkle Tree of a size that's not a power of two, all Right Subtrees are partial. At large tree sizes, that will be true most of the time, so most Right Subtrees we encounter will happen to be partial.
```
ilaril, line 439: All Right Subtrees of non-power-of-two-sized Merkle Tree being partial does not seem to hold. E.g., consider Subtree [4,6) in Merkle Tree of size 6.

- 6 is not a power of two.
- 0 <= 4 < 6 <= 6 and BIT_CEIL(6-4) = 2 divides 4, so [4,6) is a Subtree.
- 6-4=2=2^1, so [4,6) is a Full Subtree.
- [4,6) is of form [A,n), so [4,6) is Right Subtree.

What does hold is that any non-power-of-two-sized Merkle Tree has at least one Partial Right Subtree.

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -296,6 +296,12 @@ Standalone certificate:
 Landmark-relative certificate:
 : An optimized certificate containing an inclusion proof to a landmark subtree, and no signatures.
 
+Successor:
+: A Merkle Tree of size N is the successor of all Merkle Trees of size < N with which it is consistent.
+
+Predecessor:
+: A Merkle Tree of size N is the predecessor of all Merkle Trees of size > N with which it is consistent.
```
davidben, line 303: Do we actually need all this? It does seem odd to introduce this in the global terminology section but then we never actually use this term exception to explain subtrees a bit.

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -407,30 +413,44 @@ Landmark-relative certificates are constructed and used as follows. {{fig-landma
 
 # Subtrees
 
-This section extends the Merkle Tree definition in {{Section 2.1 of !RFC9162}} by defining a *subtree* of a Merkle Tree. A subtree is an interior node of a Merkle Tree, which can be efficiently shown consistent with the original Merkle Tree and any Merkle Tree with additional elements appended. This specification uses subtrees to reduce the size of inclusion proofs.
+This section extends the Merkle Tree definition in {{Section 2.1 of !RFC9162}} by defining a *subtree* of a Merkle Tree. Using subtrees allows the use of smaller inclusion proofs.
 
-## Definition of a Subtree
+A Merkle Tree is defined by an ordered list of `n` inputs, `D_n = {d[0], d[1], ..., d[n-1]}`. A subtree of a Merkle Tree consists of an interior node and all of its children. A subtree can be identified by the inputs `D_n` plus a half-open interval `[A, B)` that *exactly* identifies the entries below a single node.
```
davidben, line 418: Doesn't this bring back the original confusion? The subtree isn't necessarily an interior node of _this_ Merkle Tree. It's also a little odd in that a Merkle Tree in 9162 and 6962 is formally defined as this funky hash function. The tree only comes from implicitly looking at the call tree. (I don't know if I'd have defined it that way, but unless we want to take on redefining all that, I think that's what we have to work with.)

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -407,30 +413,44 @@ Landmark-relative certificates are constructed and used as follows. {{fig-landma
...
 * `start` is a multiple of `BIT_CEIL(end - start)`
 
 Note that, if `start` is zero, the second condition is always true.
 
-In the context of a single Merkle Tree, the subtree defined by `start` and `end` is denoted by half-open interval `[start, end)`. It contains the entries whose indices are in that half-open interval.
+## Full Subtrees
+
+Any subtree with power-of-two size is *full*, otherwise it is *partial*. Any full subtree of a given Merkle Tree is also contained in all successors to that Merkle Tree, with all hash values unchanged.
+
+## Right Subtrees
```
davidben, line 431: This terminology seems slightly unfortunate. "Right subtree" also sounds like "take the right subtree _of this node_", which means something very different.

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -407,30 +413,44 @@ Landmark-relative certificates are constructed and used as follows. {{fig-landma
...
+Partial Subtrees only occur along the right-hand edge of a Merkle Tree. In other words, all Partial Subtrees are also Right Subtrees.
 
-The *size* of the subtree is `end - start`. If the subtree's size is a power of two, it is said to be *full*, otherwise it is said to be *partial*.
+In a Merkle Tree of a size that's not a power of two, all Right Subtrees are partial. At large tree sizes, that will be true most of the time, so most Right Subtrees we encounter will happen to be partial.
 
-If a subtree is full, then it is directly contained in the tree of hash operations in `MTH(D_n)` for `n >= end`.
+A partial subtree of a given Merkle Tree will not be contained in any successors of that tree, because the successor will have different values at some of its interior nodes. However, it can efficiently be proved consistent with successors.
 
-If a subtree is partial, it is directly contained in `MTH(D_n)` only if `n = end`.
+## Covering Arbitrary Ranges Parsimoniously
```
davidben, line 443: Is the idea that we'd delete the other section then?

davidben (commented):
Sorry this slipped through my inbox. Overall, it confuses me somewhat that this PR seem to define a *ton* of concepts, but then the document overall doesn't really use them very much. Do we actually need all of these?

davidben:
WDYT about https://github.com/ietf-plants-wg/merkle-tree-certs/pull/216? It doesn't change quite as much as this PR does, but gives a hopefully clearer overview.

jsha:
Good points that this introduces more concepts than are really needed. I do like the improved intro in #216.

My goals here are:

 1. Build a better mental model of what subtrees look like and what their shape is, particularly the shape of a pair of them.
 2. Justify the constraints on subtree `start`.
 3. Be more crisp about the "contained" relation (and remove "directly contained", which I found confusing).

For (3), my feeling is that a Merkle Tree "containing" a subtree means that not only the entries are the same but also the interior node hash values. That implies that a Partial Right Subtree is only "contained" in the Merkle Tree where `n == end`. Of course, we don't expect logs to store older hash values as they evolve their tree, but it seems useful conceptually.

I'm happy to do another round of thinking and writing on this if it seems useful. I feel like I understand things a bit better now than when I wrote it.

davidben:
Hmm, yeah, the "contained" business is not great. It's especially awkward given that RFC 9162 formally defines a "tree" as a recursive function. The "tree" just comes out of drawing the call tree out. But that is very unhelpful for getting an intuition.

Yeah, if we can make it more understandable, that sounds useful! Mostly my worry with spending a ton of time on defining new concepts is that it reads as if you have to understand all these extra concepts to work with subtrees, but this spec doesn't really care as much.

(For that matter, even the existing "full" and "partial" subtrees aren't strictly necessary concepts. I don't think we actually use them, except to give some color on how the "Arbitrary Intervals" procedure behaves.)

davidben:
Forgot this was still open... I *think* the subtree rewrites sorted out the confusing bits? Is there other stuff to pull in, or should we close this out?

PR #213, opened 2026-04-20
Add log numbers to certificates and log entries

nharper:
Addresses issue #26.

### DIFF
--- a/draft-ietf-plants-merkle-tree-certs.md
+++ b/draft-ietf-plants-merkle-tree-certs.md
@@ -274,7 +274,7 @@ Monitor:
 : Parties who watch logs for certificates of interest, analogous to the role in {{Section 8.2 of ?RFC9162}}.
 
 Issuance log:
-: A log, maintained by the CA, of everything issued by that CA.
+: A log, maintained by the CA, containing certification statements issued by that CA. A CA operates some number of issuance logs, which together contain all statements issued by that CA.
 
 Cosigner:
 : A service that signs views of an issuance log, to assert correct operation and other properties about the entries.
@@ -893,29 +893,25 @@ A snapshot of the log is known as a *checkpoint*. A checkpoint is identified by 
 
 Cosigners ({{cosigners}}) sign assertions about the state of the issuance log. A Merkle Tree CA operates a combination of an issuance log and CA cosigner ({{certification-authority-cosigners}}) that authenticates the log state and certifies the contents. External cosigners may also be deployed to assert correct log operation or provide other services to relying parties ({{trusted-cosigners}}).
 
-## Log Parameters
+## Relationship with Certification Authorities {#ca-ids}
 
-An issuance log has the following parameters:
+Each issuance log is run by a Certification Authority. A CA can run multiple related issuance logs. Each log is assigned a positive integer to identify it, and these log numbers MUST be assigned in increasing order.
 
-* A log ID, which uniquely identifies the log. See {{log-ids}}.
-* A collision-resistant cryptographic hash function. SHA-256 {{!SHS=DOI.10.6028/NIST.FIPS.180-4}} is RECOMMENDED.
-* A minimum index, which is the index of the first log entry which is available. See {{log-pruning}}. This value changes over the lifetime of the log.
+Each Merkle Tree Certificate CA has a *CA ID* to identify it. This CA ID is a trust anchor ID {{!I-D.ietf-tls-trust-anchor-ids}}, and its entire object identifier (OID) arc is allocated for the issuance logs run by that CA.
 
-Throughout this document, the hash algorithm in use is referred to as HASH, and the size of its output in bytes is referred to as HASH_SIZE.
+An individual issuance log run by a CA can be identified by a *log ID* constructed from the *CA ID* and log number. The log ID is also a trust anchor ID.
 
-## Log IDs
+OIDs under this arc are allocated by the Merkle Tree Certificates protocol. Given a CA ID whose OID representation is `caID` and a log operated by that CA with log number `N`, this document allocates the following OIDs:
 
-Each issuance log is identified by a *log ID*, which is a trust anchor ID {{!I-D.ietf-tls-trust-anchor-ids}}.
+* The OID `{caID log(0) N}` represents the log ID for log number `N`.
 
-When allocating log IDs, the entire object identifier (OID) arc is allocated for use with the issuance log. OIDs under this arc are allocated by the Merkle Tree Certificates protocol. Given a log ID whose OID representation is `logID`, this document allocates the following OIDs:
+* For each non-negative integer `L`, the OID `{caID landmarks(1) N L}` represents the landmark ({{landmark-tree-sizes}}) with number `L`. These OIDs may used as a trust anchor ID, as described in {{landmark-relative-certificates-tls}}. These OIDs are used when it is necessary to identify an individual landmark, e.g. as in the retry mechanism described {{Section 4.3 of !I-D.ietf-tls-trust-anchor-ids}}.
 
-* For each non-negative integer `L`, the OID `{logID landmarks(0) L}` represents the landmark ({{landmark-tree-sizes}}) with number `L`. These OIDs may used as a trust anchor ID, as described in {{landmark-relative-certificates-tls}}. These OIDs are used when it is necessary to identify an individual landmark, e.g. as in the retry mechanism described {{Section 4.3 of !I-D.ietf-tls-trust-anchor-ids}}.
-
-* For each non-negative integer `L`, the OID `{logID landmarkGroups(1) L}` represents a trust anchor group ({{Section 5 of !I-D.ietf-tls-trust-anchor-ids}}) containing landmark number `L` and earlier landmarks, as defined in {{single-log-landmark-groups}}. These OIDs may be used to advertise a series of landmarks at once.
+* For each non-negative integer `L`, the OID `{caID landmarkGroups(2) N L}` represents a trust anchor group ({{Section 5 of !I-D.ietf-tls-trust-anchor-ids}}) containing landmark number `L` and earlier landmarks, as defined in {{single-log-landmark-groups}}. These OIDs may be used to advertise a series of landmarks at once.
 
 Future extensions to this protocol MAY define further allocations.
 
-An issuance log's log ID determines a PKIX distinguished name ({{Section 4.1.2.4 of !RFC5280}}). The distinguished name has a single relative distinguished name, which has a single attribute. The attribute has type `id-rdna-trustAnchorID`, defined below:
+A CA ID determines a PKIX distinguished name ({{Section 4.1.2.4 of !RFC5280}}) that can be used in the issuer or subject field of an X.509 TBSCertificate. This distinguished name has a single relative distinguished name, which has a single attribute. The attribute has type `id-rdna-trustAnchorID`, defined below:
 
 ~~~asn.1
 id-rdna-trustAnchorID OBJECT IDENTIFIER ::= {
@@ -923,7 +919,7 @@ id-rdna-trustAnchorID OBJECT IDENTIFIER ::= {
     mechanisms(5) pkix(7) rdna(25) TBD }
 ~~~
 
-The attribute's value is a RELATIVE-OID containing the trust anchor ID's ASN.1 representation. For example, the distinguished name for a log named `32473.1` would be represented in syntax of {{?RFC4514}} as:
+The attribute's value is a RELATIVE-OID containing the trust anchor ID's ASN.1 representation. For example, the distinguished name for a CA with ID `32473.1` would be represented in syntax of {{?RFC4514}} as:
 
 ~~~
 1.3.6.1.5.5.7.25.TBD=#0d0481fd5901
@@ -931,15 +927,25 @@ The attribute's value is a RELATIVE-OID containing the trust anchor ID's ASN.1 r
 
 For initial experimentation, early implementations of this design will:
 
-1. Use UTF8String to represent the attribute's value rather than RELATIVE-OID. The UTF8String contains trust anchor ID's ASCII representation, e.g. `324731.1`.
+1. Use UTF8String to represent the attribute's value rather than RELATIVE-OID. The UTF8String contains trust anchor ID's ASCII representation, e.g. `32473.1`.
 
 1. Use the OID 1.3.6.1.4.1.44363.47.1 instead of `id-rdna-trustAnchorID`.
 
-For example, the distinguished name for a log named `32473.1` would be represented in syntax of {{?RFC4514}} as:
+For example, the distinguished name for a CA with ID `32473.1` would be represented in syntax of {{?RFC4514}} as:
 
 ~~~
 1.3.6.1.4.1.44363.47.1=#0c0733323437332e31
 ~~~
+
+## Log Parameters
+
+An issuance log has the following parameters:
+
+* A log ID, constructed from the CA ID and log number, which uniquely identifies the log. See {{ca-ids}}.
+* A collision-resistant cryptographic hash function. SHA-256 {{!SHS=DOI.10.6028/NIST.FIPS.180-4}} is RECOMMENDED. The same hash function MUST be used for all logs under the same CA ID.
+* A minimum index, which is the index of the first log entry which is available. See {{log-pruning}}. This value changes over the lifetime of the log.
+
+Throughout this document, the hash algorithm in use is referred to as HASH, and the size of its output in bytes is referred to as HASH_SIZE.
 
 ## Log Entries
 
@@ -962,7 +968,7 @@ struct {
 } MerkleTreeCertEntry;
 ~~~
 
-When `type` is `null_entry`, the entry does not represent any information. The entry at index zero of every issuance log MUST be of type `null_entry`. This avoids zero serial numbers in the certificate format ({{certificate-format}}). Other entries MAY have type `null_entry`.
+When `type` is `null_entry`, the entry does not represent any information. Entries at any index in the log MAY have type `null_entry`.
 
 When `type` is `tbs_cert_entry`, `N` is the number of bytes needed to consume the rest of the input. A MerkleTreeCertEntry is expected to be decoded in contexts where the total length of the entry is known.
 
@@ -988,7 +994,7 @@ The fields of a TBSCertificateLogEntry are defined as follows:
 
 * `version`, `validity`, `subject`, `issuerUniqueID`, `subjectUniqueID`, and `extensions` have the same semantics as the corresponding TBSCertificate fields, defined in {{Section 4.1.2 of !RFC5280}}.
 
-*  `issuer` is the issuance log's log ID as a PKIX distinguished name, as described in {{log-ids}}.
+*  `issuer` is the CA ID as a PKIX distinguished name, as described in {{ca-ids}}.
 
 * `subjectPublicKeyAlgorithm` describes the algorithm of the subject's public key. It is constructed identically to the `algorithm` field of a SubjectPublicKeyInfo ({{Section 4.1.2.7 of !RFC5280}}).
 
@@ -1034,7 +1040,7 @@ This signature format is designed to be compatible with the ML-DSA-44 signature 
 
 `label` is a fixed prefix for domain separation. Its value MUST be the string `subtree/v1`, followed by a newline (U+000A), followed by a zero byte (U+0000).
 
-`cosigner_name` and `log_origin` are computed from the cosigner ID and the issuance log's ID ({{log-ids}}), respectively. They contain the concatenation of:
+`cosigner_name` and `log_origin` are computed from the cosigner ID and the issuance log's ID ({{ca-ids}}), respectively. They contain the concatenation of:
 
 * The 16-byte ASCII string `oid/1.3.6.1.4.1.`
 * The trust anchor ID's ASCII representation ({{Section 3 of !I-D.ietf-tls-trust-anchor-ids}})
@@ -1079,7 +1085,7 @@ Other documents or deployments MAY define other signature schemes and formats. L
 
 ## Certification Authority Cosigners
 
-A *CA cosigner* is a cosigner ({{cosigners}}) that certifies the contents of a log. Each CA MUST operate a CA cosigner whose cosigner ID is the same as the log ID ({{log-ids}}). A CA cosigner MUST NOT sign checkpoints or subtrees for logs other than its corresponding issuance log.
+A *CA cosigner* is a cosigner ({{cosigners}}) that certifies the contents of a log. Each CA MUST operate a CA cosigner whose cosigner ID is the same as its CA ID ({{ca-ids}}). A CA cosigner MUST NOT sign checkpoints or subtrees for logs other than its corresponding issuance log.
 
 When a CA cosigner signs a subtree, it makes the additional statement that it has certified each entry in the subtree. For example, a domain-validating CA states that it has performed domain validation for each entry, at some time consistent with the entry's validity dates. CAs are held responsible for every entry in every subtree they sign. Proving an entry is included ({{subtree-inclusion-proofs}}) in a CA-signed subtree is sufficient to prove the CA certified it.
 
@@ -1194,9 +1200,9 @@ For any given TBSCertificateLogEntry, there are multiple possible certificates t
 
 The information is encoded in an X.509 Certificate {{!RFC5280}} as follows:
 
-The TBSCertificate's `version`, `issuer`, `validity`, `subject`, `issuerUniqueID`, `subjectUniqueID`, and `extensions` MUST be equal to the corresponding fields of the TBSCertificateLogEntry. If any of `issuerUniqueID`, `subjectUniqueID`, or `extensions` is absent in the TBSCertificateLogEntry, the corresponding field MUST be absent in the TBSCertificate. Per {{log-entries}}, this means `issuer` MUST be the issuance log's log ID as a PKIX distinguished name, as described in {{log-ids}}.
+The TBSCertificate's `version`, `issuer`, `validity`, `subject`, `issuerUniqueID`, `subjectUniqueID`, and `extensions` MUST be equal to the corresponding fields of the TBSCertificateLogEntry. If any of `issuerUniqueID`, `subjectUniqueID`, or `extensions` is absent in the TBSCertificateLogEntry, the corresponding field MUST be absent in the TBSCertificate. Per {{log-entries}}, this means `issuer` MUST be the issuance log's CA ID as a PKIX distinguished name, as described in {{ca-ids}}.
 
-The TBSCertificate's `serialNumber` MUST contain the zero-based index of the TBSCertificateLogEntry in the log. {{Section 4.1.2.2 of !RFC5280}} forbids zero as a serial number, but {{log-entries}} reserves entry zero with a `null_entry` type, so the index will be positive. This encoding is intended to avoid implementation errors by having the serial numbers and indices off by one.
+The TBSCertificate's `serialNumber` is constructed from the zero-based index of the TBSCertificateLogEntry in the log and the log's number ({{ca-ids}}). The `serialNumber` MUST be equal to `(log_number << 64) | index`.
 
 The TBSCertificate's `subjectPublicKeyInfo` contains the specified public key. Its `algorithm` field MUST match the TBSCertificateLogEntry's `subjectPublicKeyAlgorithm`. Its hash MUST match the TBSCertificateLogEntry's `subjectPublicKeyInfoHash`.
 
@@ -1322,15 +1328,15 @@ Proof sizes grow logarithmically, so 32 hashes, or 1024 bytes, is sufficient for
 
 ## Representing Certification Authorities
 
-This section defines the X.509 Certificate {{!RFC5280}} representation of a Merkle Tree Certificate CA. It identifies the issuance log ({{issuance-logs}}) and the associated CA cosigner ({{certification-authority-cosigners}}). This information is encoded as follows:
+This section defines the X.509 Certificate {{!RFC5280}} representation of a Merkle Tree Certificate CA. It identifies the CA cosigner ({{certification-authority-cosigners}}) and associated issuance logs. This information is encoded as follows:
 
-* The `subject` field MUST be the CA issuance log's log ID as a PKIX distinguished name, as described in {{log-ids}}.
+* The `subject` field MUST be the CA ID as a PKIX distinguished name, as described in {{ca-ids}}.
 
 * The `subjectPublicKeyInfo` field MUST be the public key of the CA cosigner {{certification-authority-cosigners}}.
 
 * The `extensions` field MUST contain a critical extension of type id-pe-mtcCertificationAuthority, defined below.
 
-* The subject key identifier extension ({{Section 4.2.1.2 of !RFC5280}}), if present, SHOULD be set to CA's log ID {{log-ids}}. The log ID is encoded in its binary representation, as defined in {{Section 3 of !I-D.ietf-tls-trust-anchor-ids}}.
+* The subject key identifier extension ({{Section 4.2.1.2 of !RFC5280}}), if present, SHOULD be set to the CA ID {{ca-ids}}. The CA ID is encoded in its binary representation, as defined in {{Section 3 of !I-D.ietf-tls-trust-anchor-ids}}.
 
 Other fields and extensions in {{!RFC5280}} apply unmodified. In particular:
 
@@ -1365,11 +1371,11 @@ For initial experimentation, early implementations of this design will use the O
 
 The fields of a MTCCertificationAuthority structure are defined as follows:
 
-* `logHash` describes the hash algorithm used in the log. For example, if the hash is SHA-256, it would be `mda-sha256` as defined in {{Section 8 of !RFC5912}}.
+* `logHash` describes the hash algorithm used by all logs operated by this CA. For example, if the hash is SHA-256, it would be `mda-sha256` as defined in {{Section 8 of !RFC5912}}.
 
 * `sigAlg` is the CA cosigner's signature algorithm ({{signature-algorithms}}).
 
-* `minSerial` is an integer describing the minimum allowed serial number (entry index) in the issuance log ({{log-pruning}}).
+* `minSerial` is an integer describing the minimum allowed serial number from this CA. Since the serial number encodes both the log number ({{ca-ids}}) and the entry index into a specific log, it can be used to set a minimum allowed log number or a minimum allowed index in a particular log ({{log-pruning}}).
 
 If this extension is present, the key described in `subjectPublicKeyInfo` is a CA cosigner key and subject to the usage restrictions described in {{certification-authority-cosigners}}. In particular, it MUST NOT be used to directly sign TBSCertificate structures.
 
@@ -1385,24 +1391,30 @@ This section discusses how relying parties verify Merkle Tree Certificates.
 
 In order to accept certificates from a Merkle Tree CA, a relying party MUST be configured with:
 
-* The log ID ({{log-ids}})
-* The log hash algorithm, e.g. SHA-256
+* The CA's ID ({{ca-ids}})
+* The hash algorithm used in each log, e.g. SHA-256
 * The CA cosigner, and any other supported cosigners, as pairs of cosigner ID and public key
 * A policy on which combinations of cosigners to accept in a certificate ({{trusted-cosigners}})
-* An optional list of trusted subtrees, with their hashes, that are known to be consistent with the relying party's cosigner requirements ({{trusted-subtrees}})
-* A list of revoked ranges of indices ({{revocation-by-index}})
+* A list of revoked ranges of serial numbers ({{revocation-by-index}})
+
+A relying party MAY be configured on a per-CA basis with a minimum log number, such that it will only trust certificates from logs with a log number greater than or equal to that minimum.
+
+Additionally, a relying party MAY be configured on a per-log basis with the following:
+
+* The log's log number ({{ca-ids}})
+* A list of trusted subtrees, with their hashes, that are known to be consistent with the relying party's cosigner requirements ({{trusted-subtrees}})
 
 This information may be obtained from a CA certificate structure, defined in {{representing-certification-authorities}}:
 
-* The log ID is determined from the certificate's subject.
+* The CA ID is determined from the certificate's subject.
 
 * The log hash algorithm is determined from the id-pe-mtcCertificationAuthority extension.
 
-* The CA cosigner is determined from the certificate's subject public key and id-pe-mtcCertificationAuthority extension. The CA's cosigner ID is the log ID, determined above. The relying party incorporates this cosigner into its cosigner policy based on the guidance in {{trusted-cosigners}}.
+* The CA cosigner is determined from the certificate's subject public key and id-pe-mtcCertificationAuthority extension. The CA's cosigner ID is the same as its CA ID. The relying party incorporates this cosigner into its cosigner policy based on the guidance in {{trusted-cosigners}}.
 
 * No trusted subtrees are directly represented by the CA certificate structure, but the relying party MAY incorporate trusted subtrees from out-of-band information.
 
-* The revoked index ranges include the half-open range `[0, minSerial)`, but the relying party MAY incorporate additional ranges from out-of-band information.
+* The revoked serial number ranges include the half-open range `[0, minSerial)`, but the relying party MAY incorporate additional ranges from out-of-band information.
 
 ## Verifying Certificate Signatures
 
@@ -1412,7 +1424,16 @@ When verifying the signature of an X.509 certificate (Step (a)(1) of {{Section 6
 
 1. Decode the `signatureValue` as an MTCProof, as described in {{certificate-format}}.
 
-1. Let `index` be the certificate's serial number. If `index` is contained in one of the relying party's revoked ranges ({{revocation-by-index}}), abort this process and fail verification.
+1. Let `serial` be the certificate's serial number. If `serial` is negative, abort this process and fail verification.
+
+1. Let `index` be the least significant 64 bits of `serial` and let `log_number` be `serial >> 64`.
+
+1. Let `log_id` be the log ID constructed from the CA ID in `issuer` and the `log_number` ({{ca-ids}})
+
+1. If any of the following conditions are true, abort this process and fail verification:
+   * The `log_number` is zero
+   * The `log_number` is smaller than the minimum log number for that CA ({{relying-party-configuration}})
+   * `serial` is contained in one of the relying party's revoked ranges ({{revocation-by-index}})
 
 1. Construct a TBSCertificateLogEntry as follows:
    1. Copy the `version`, `issuer`, `validity`, `subject`, `issuerUniqueID`, `subjectUniqueID`, and `extensions` fields from the TBSCertificate.
@@ -1423,7 +1444,7 @@ When verifying the signature of an X.509 certificate (Step (a)(1) of {{Section 6
 
 1. Let `expected_subtree_hash` be the result of evaluating the MTCProof's `inclusion_proof` for entry `index`, with hash `entry_hash`, of the subtree described by the MTCProof's `start` and `end`, following the procedure in {{evaluating-a-subtree-inclusion-proof}}. If evaluation fails, abort this process and fail verification.
 
-1. If `[start, end)` matches a trusted subtree ({{trusted-subtrees}}), check that `expected_subtree_hash` is equal to the trusted subtree's hash. Return success if it matches and failure if it does not.
+1. If `[start, end)` matches a trusted subtree ({{trusted-subtrees}}) for the log identified by `log_id` ({{ca-ids}}), check that `expected_subtree_hash` is equal to the trusted subtree's hash. Return success if it matches and failure if it does not.
 
 1. Otherwise, check that the MTCProof's `signatures` contain a sufficient set of valid signatures from cosigners to satisfy the relying party's cosigner requirements ({{trusted-cosigners}}). Unrecognized cosigners MUST be ignored. Signatures are verified as described in {{signature-format}}. Reconstruct the CosignedMessage from MTCProof's `start` and `end`, the cosigner ID for `cosigner_name`, the log ID for `log_origin`, `expected_subtree_hash` for `subtree_hash`, and `timestamp` set to zero.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -944,6 +951,7 @@ When `type` is `tbs_cert_entry`, `N` is the number of bytes needed to consume th
 ~~~asn.1
 TBSCertificateLogEntry ::= SEQUENCE {
     version               [0] EXPLICIT Version DEFAULT v1,
+    logNumber                 INTEGER,
```
davidben: Do we need the log number in the log entries? I _think_ we actually don't. (Really we don't need the issuer in the log entries either. Only reason I put it in there was that X.509 name parsing is a trash fire and it seemed easier to fix it into the entry than to worry about whether we were accidentally admitting some kind of malleability.)

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -992,7 +1002,9 @@ opaque HashValue[HASH_SIZE];
 opaque TrustAnchorID<1..2^8-1>;
 
 struct {
-    TrustAnchorID log_id;
+    TrustAnchorID ca_id;
+    /* TODO: Are we ok with capping log numbers at 2^64-1? */
```
davidben: I am *extremely* OK with capping log numbers hehehe.

That said, after https://github.com/C2SP/C2SP/pull/237, we'll likely want to say that logs have IDs, constructed by appending the number to the CA ID, and that the number is encoded via `log_id` instead of as a separate field. Though doing it this way works too. (Though given that Appendix C.1 already basically depends on a notion of a "log ID" maybe we should just go ahead and get that defined?)

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1453,7 +1475,7 @@ In applications that use additional trust anchor ranges, relying parties MAY sen
 
 ## Using Trust Anchor IDs
 
-A standalone certificate will generally be accepted by relying parties that trust the issuing CA. To determine this, a standalone certificate has a trust anchor ID of the corresponding log ID ({{log-ids}}). The authenticating party can obtain this information either by parsing the certificate's issuer field or via out-of-band information as described in {{Section 3.2 of !I-D.ietf-tls-trust-anchor-ids}}. Authenticating and relying parties SHOULD use the `trust_anchors` extension to determine whether the standalone certificate would be acceptable.
+A standalone certificate will generally be accepted by relying parties that trust the issuing CA. To determine this, a standalone certificate has a trust anchor ID of the corresponding log ID ({{ca-ids}}). The authenticating party can obtain this information either by parsing the certificate's issuer field or via out-of-band information as described in {{Section 3.2 of !I-D.ietf-tls-trust-anchor-ids}}. Authenticating and relying parties SHOULD use the `trust_anchors` extension to determine whether the standalone certificate would be acceptable.
```
davidben: Nit: This still says log ID, but I think you mean issuer ID now?

draft-ietf-plants-merkle-tree-certs.md
davidben: Oh, we should remove the requirement that index zero is `null_entry` if we're going to force it positive by way of log number. And maybe replace it with some other justification for null entries. Perhaps "Null entries MAY optionally be used as part of periodic health checks, but otherwise do not certify any information."

I don't feel strongly about which way we force it. When I'd thought about it, I'd leaned towards allowing log number zero, but I don't think I came up with a clear reason either way. Some (very very weak) points I can think of:

* Forcing it by log number _does_ make your very first log's certificates ever-so-slightly shorter, but as soon as you have to rotate it once, that'll change, so this is probably not worth optimizing for.

* We have to answer dumb questions like "if we get a certificate whose log number part is zero, must we reject the signature"? Although we have the corresponding dumb question for the indices too, we just (currently) get to punt it to the serial number situation.

* A zero log number also feels like a more natural "no-op" version of log generations, if your application would like to have nothing to do with this stuff.

* Using `null_entry` means we burn entry zero of _every_ log, unless we decide to special case log zero, which seems kinda odd.

* Using the log number means we don't need to define `null_entry` at all. But it seems vaguely useful for health checks and to exercise the entry type out the gate, so maybe we want that anyway.

🤷  I think I'm happy with whichever.

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -259,7 +259,7 @@ Monitor:
 : Parties who watch logs for certificates of interest, analogous to the role in {{Section 8.2 of ?RFC9162}}.
 
 Issuance log:
-: A log, maintained by the CA, of everything issued by that CA.
+: A log, maintained by the CA, of things issued by that CA. A CA operates some number of issuance logs, which together contain everything issued by that CA.
```
davidben: Nit: "Things" is a bit vague. :-) It's arguably an existing issue ("everything") but "things" somehow feels more egregious to me.

What about...

```suggestion
: A log, maintained by the CA, containing information certified by that CA. A CA operates some number of issuance logs, which together contain all information issued by that CA.
```
(I toyed around with "containing some of the CA's issuance events", but then we have to define an "issuance event" and that seemed a pain.)

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -874,21 +874,9 @@ A snapshot of the log is known as a *checkpoint*. A checkpoint is identified by
...
-An issuance log has the following parameters:
-
-* A log ID, which uniquely identifies the log. See {{log-ids}}.
-* A collision-resistant cryptographic hash function. SHA-256 {{!SHS=DOI.10.6028/NIST.FIPS.180-4}} is RECOMMENDED.
-* A minimum index, which is the index of the first log entry which is available. See {{log-pruning}}. This value changes over the lifetime of the log.
-
-Throughout this document, the hash algorithm in use is referred to as HASH, and the size of its output in bytes is referred to as HASH_SIZE.
-
-## Log IDs
+## Certification Authority Identifiers {#ca-ids}
```
davidben: Nit: It seems in prose you call this an "issuer ID" but the section is called a CA ID.

We do want to avoid confusing with "CA cosigner ID", so long as a CA can have multiple cosigners, so I think I like sticking _some_ word like "issuer" in there somehow. Though "CA issuer" is also a little awkward because it might mean the issuer _of_ the CA, rather than the CA _as_ an issuer.

I'm not sure what would be good names. I'm not super satisfied with any of them, but I don't have a great suggestion...

Spitballing, I haven't thought about the implications: what if the "issuance log"  was all the log numbers together and made up of a bunch of trees. And then each log number was actually a "log instance number" which referred to one "log instance" within the "log". Or perhaps a "tree number" and referred to a specific "issuance tree" within the "issuance log".

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -914,6 +902,25 @@ For example, the distinguished name for a log named `32473.1` would be represent
...
+* A CA issuer ID. See {{ca-ids}}.
+* A positive log number. The CA issuer ID and log number uniquely identify the log; see {{log-ids}}.
+* A collision-resistant cryptographic hash function. SHA-256 {{!SHS=DOI.10.6028/NIST.FIPS.180-4}} is RECOMMENDED.
+* A minimum index, which is the index of the first log entry which is available. See {{log-pruning}}. This value changes over the lifetime of the log.
+
+Throughout this document, the hash algorithm in use is referred to as HASH, and the size of its output in bytes is referred to as HASH_SIZE.
+
+## Identifying Issuance Logs {#log-ids}
+
+A CA can operate one or more issuance logs. Each log that the CA operates is assigned a log number, which is a positive 64-bit integer. Log numbers must be unique per CA.
```
davidben: Super nitpicky nitpick: it may be worth saying "positive 64-bit integer, greater than zero", just to make sure people realize it is indeed non-zero. I think "positive integer" pretty consistently means non-zero, but "natural number" is a trainwreck.

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1147,9 +1159,11 @@ For any given TBSCertificateLogEntry, there are multiple possible certificates t
...
 
-The TBSCertificate's `serialNumber` MUST contain the zero-based index of the TBSCertificateLogEntry in the log. {{Section 4.1.2.2 of !RFC5280}} forbids zero as a serial number, but {{log-entries}} reserves entry zero with a `null_entry` type, so the index will be positive. This encoding is intended to avoid implementation errors by having the serial numbers and indices off by one.
+The TBSCertificate's `serialNumber` is constructed from the zero-based index of the TBSCertificateLogEntry in the log and the logNumber in the TBSCertificateLogEntry. The `serialNumber` MUST be equal to `(logNumber << 64) | index`, i.e. the logNumber from the TBSCertificateLogEntry bit-shifted left 64 bits and bitwise ORed with the zero-based index of the log entry.
```
davidben: Nit: We already define `<<` and `|`  in the notation section. Realistically, I think anyone who knows what "bit-shifted left" and "bitwise ORed" means will also know what those symbols mean. It's just very wordy. If we want to clarify it, perhaps, `i.e. logNumber * 2<sup>64</sup> + index` or something.

davidben (commented):
Thanks! It's a bit late on the east coast, so I haven't looked over everything yet, but I figure I may as well flush initial comments now.

draft-ietf-plants-merkle-tree-certs.md
nharper: Didn't realize those are already in the notation section. Removed the prose.

draft-ietf-plants-merkle-tree-certs.md
nharper: Clarified "greater than zero".

draft-ietf-plants-merkle-tree-certs.md
nharper: Reworded. I wanted to `s/things/certificates/`, but they're not certificates, so I went with "certification statements".

draft-ietf-plants-merkle-tree-certs.md
nharper: I removed the requirement, but kept `null_entry` around, because it seems useful, especially to exercise the entry type. (Perhaps log operators might find it useful to pad out trees to a certain size when generating checkpoints or landmarks.) I don't think we should special case log zero (I don't think log zero should exist).

draft-ietf-plants-merkle-tree-certs.md
nharper: I think so too, changed that.

draft-ietf-plants-merkle-tree-certs.md
nharper: Removed the TODO. I have a bit of a preference for having this draft not be opinionated on OID structure, which means that we shouldn't be constructing OIDs in it. Even if we keep the log number separate from the OID, defining a "log ID" that combines the two (and has whatever serialization formats we need) is probably useful. I'll follow up with that.

draft-ietf-plants-merkle-tree-certs.md
nharper: For the same reason that we probably don't need the issuer, we probably don't need the log number either. I haven't thought through whether removing it could cause issues down the line. I'm going to leave this comment thread open for now and I'll probably remove it later.

Bren2010:
We had a decent discussion about this on the list, but to restate my objections to this approach:

- If the cosigners in an MTC deployment are **witnesses**, then there's no strong incentive to keep all witnesses fully up-to-date all the time since updating them is cheap. This could lead to a situation where one witness is far behind the others and then is suddenly updated with a large number of certificates. If that push is corrupted in some way, it can create a fork far back in the log. Revoking by index in this scenario would likely cause an unacceptable amount of collateral damage. That leaves you with a fork full of certificates that are signed, unrevoked, and not visible to monitors, which is a bad outcome.

- If the cosigners in an MTC deployment are **mirrors**, then it's less likely that one mirror would be allowed to fall substantially behind the others and you have the additional benefit that each mirror can see all the certificates on both sides of the fork. Because of that, it's simpler to just ask a mirror that falls onto a fork to "jump" back to the correct fork (where part of "jumping" involves verifying that no certificates were removed, just potentially reordered). I say that's simpler because:
  - It requires no protocol change.
  - It allows monitors to know that they've observed all issued certificates if they have a quorum of cosigners on whichever fork they're on. (To contrast that with this approach: this approach says that if a mirror is forked, it remains forked forever. Which means, to observe all certificates in all forks, we need to sync with *all* mirrors.)

davidben:
Happy to consider alternate approaches that meet both the operational and security requirements of running a real-world PKI. This one does and I've yet to see any others that do.

> If the cosigners in an MTC deployment are witnesses, then there's no strong incentive to keep all witnesses fully up-to-date all the time since updating them is cheap. This could lead to a situation where one witness is far behind the others and then is suddenly updated with a large number of certificates. If that push is corrupted in some way, it can create a fork far back in the log. Revoking by index in this scenario would likely cause an unacceptable amount of collateral damage. That leaves you with a fork full of certificates that are signed, unrevoked, and not visible to monitors, which is a bad outcome.

So, your argument is saying witness-based transparency systems do not work, and so we don't need to consider them. That is quite a bold claim! This claim does not seem to track. At most we might want to refine the revocation setup for transparent PKIs a bit.

First, the assertion that there's no incentive to keep witnesses up-to-date does not make sense. _Any_ party in the ecosystem is capable of computing the values needed to catch some witness up. Indeed, in so far as this doomsday revocation situation is a real concern, that in itself provides incentives for the ecosystem to keep them up-to-date. Remember it only takes *one* party to care and the witnesses will catch up.

Now, for this scenario to happen, the monitoring ecosystem must have been fine with a lagging witness, i.e. because it managed to get an all-quorums-intersecting view of the log from all the other witnesses. And because it has been quite some time, all the certificates in that view were clearly visible to the monitors, or they would have sounded a different alarm.

That means, in the scenario you describe, it is clear that this one witness's view is the one to be revoked. We probably should say something about being able to close off entries at a (witness, log) basis rather than log-wide, but that's a pretty minor change.

But also this scenario just won't happen. Again, the premise that no one has an incentive for up-to-date witnesses is simply false. If a PKI such that a single weeks-behind witness is a risk of a weeks-long revocation event, that PKI should actively try to get that witness caught up, and raise the alarm when it cannot.

> If the cosigners in an MTC deployment are mirrors, then it's less likely that one mirror would be allowed to fall substantially behind the others and you have the additional benefit that each mirror can see all the certificates on both sides of the fork. Because of that, it's simpler to just ask a mirror that falls onto a fork to "jump" back to the correct fork (where part of "jumping" involves verifying that no certificates were removed, just potentially reordered). I say that's simpler because:

So, first off, it doesn't really matter whether we can save mirrors from this fate if we also want to support witness-based PKIs. I expect the Web PKI to use mirrors (it's much easier to reason about), but I think it's valuable for MTC to work for other applications.

Anyway, all this presumes a mirror design that cannot get wedged by missing preimages. However, the [current mirror protocol](https://c2sp.org/tlog-mirror) meets its scalability and consistency requirements by operating an internal witness to pre-commit to entries before they are uploaded. Once you have that, all the failure modes of a witness carry over.

Is that the right mirror design? I don't know. Feedback on that one would be great. But, whatever the design is, we need something that can operate at the scale needed for the Web PKI. That design was written to fit into the current CT log architecture, which has been proven to work.

Given the operational risks here and PQ transition timelines, I think deviating from the CT ecosystem's existing operational experience is a very, _very_ tall order. Log numbers reduce our operational risk and switching logs matches existing CT playbooks, while still providing all the tools needed to meet transparency goals. If N years later, we figure out how to not need them, great. RPs start enforcing a maximum log number of 1 and we all move on. But, in a vacuum, the bar is pretty high to argue "oh these failure modes are impossible" or "oh no this extra synchronizing operation will scale".

(But if you think you have a different mirror design that can scalably support all these operations, *including your proposed reconciliation operation*, great! I'd encourage you to work with the CT log operator community on this. That doesn't remove the need for this change.)

Bren2010:
> First, the assertion that there's no incentive to keep witnesses up-to-date does not make sense. Any party in the ecosystem is capable of computing the values needed to catch some witness up.

The document doesn't specify the protocol between the witness and the CA, but it does say "A cosigner MAY expose a private interface for the CA, to reduce denial-of-service risk" and also mentions the idea of there being witnesses that the CA doesn't communicate with at all. I understand your point as being that it's possible to avoid this happening, but my point is that it's a very big footgun that's non-obvious and seems to be consistent with the text of the document.

> Anyway, all this presumes a mirror design that cannot get wedged by missing preimages. However, the [current mirror protocol](https://c2sp.org/tlog-mirror) meets its scalability and consistency requirements by operating an internal witness to pre-commit to entries before they are uploaded. Once you have that, all the failure modes of a witness carry over.

My understanding is that the tlog-mirror doesn't produce a signature over a portion of the log until it has durably persisted all of the log entries. That's the only guarantee that we need for this to work. Support for a reconciliation operation can be added to tlog-mirror in a backwards-compatible way, and without affecting anything about how MTC works.

> If N years later, we figure out how to not need them, great. RPs start enforcing a maximum log number of 1 and we all move on. But, in a vacuum, the bar is pretty high to argue "oh these failure modes are impossible" or "oh no this extra synchronizing operation will scale".

The idea that this can be removed later is only true from the perspective of an individual browser. Monitors need to care about what a long tail of browsers might accept. So once this is added and some swath of versions of browsers support it, monitors will need to support it for as long as any of those browser-versions have non-negligible usage. If you don't add this feature now, you won't have to slowly phase it out of the ecosystem later.

draft-ietf-plants-merkle-tree-certs.md
davidben: Well, the considerations around the issuer are mostly malleability of the TBSCertificate, whether it's possible that a different issuer byte string might result in the same TBSCertificateLogEntry. They can't if the parser

I don't think that directly applies to the logNumber because we're already having to decode and reencode a different value when we go from TBSCertificate to TBSCertificateLogEntry. If we put the whole serial number in, that could help, but that makes the log API very complex because you need to synchronize generating the log entry bytes. So, for serial number, [security considerations](https://ietf-plants-wg.github.io/merkle-tree-certs/draft-ietf-plants-merkle-tree-certs.html#name-certificate-malleability) talk about needing to be strict in processing that.

We *could* have done the same with the issuer field, but since X.509 names are tied up with canonicalization, I opted to just copy it in and not think too hard about it.

But since none of that applies to `logNumber`, I think that points to omitting it.

draft-ietf-plants-merkle-tree-certs.md
davidben: https://github.com/ietf-plants-wg/merkle-tree-certs/pull/215 is uploaded now and requires a log ID.

If there is a log ID, we need to be *at least* opinionated enough to be able to deterministically construct the signature payload. The point of this scheme is that your one config blob will get you all the log numbers at once

We *could* potentially be unopinionated about things like cosigner IDs or landmark IDs (but I think it is better to be opinionated). That's because, in the unopinionated model, those would be a couple of extra fields in your config.

(The landmark IDs are themselves already constructed OIDs because they take up a whole OID arc. That is fine and the reason we used OIDs.)

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1147,9 +1158,11 @@ For any given TBSCertificateLogEntry, there are multiple possible certificates t
...
 
-The TBSCertificate's `serialNumber` MUST contain the zero-based index of the TBSCertificateLogEntry in the log. {{Section 4.1.2.2 of !RFC5280}} forbids zero as a serial number, but {{log-entries}} reserves entry zero with a `null_entry` type, so the index will be positive. This encoding is intended to avoid implementation errors by having the serial numbers and indices off by one.
+The TBSCertificate's `serialNumber` is constructed from the zero-based index of the TBSCertificateLogEntry in the log and the logNumber in the TBSCertificateLogEntry. The `serialNumber` MUST be equal to `(logNumber << 64) | index`.
+
+TODO: {{log-entries}} reserves entry zero with a `null_entry` type so the index will be positive. This was done because {{Section 4.1.2.2 of !RFC5280}} forbids zero as a serial number. However, with the inclusion of a positive `logNumber` in the `serialNumber`, the serial number will always be positive regardless of the index, and the `null_entry` type could be removed.
```
davidben: Out-of-date TODO?

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -914,6 +902,25 @@ For example, the distinguished name for a log named `32473.1` would be represent
...
+
+Throughout this document, the hash algorithm in use is referred to as HASH, and the size of its output in bytes is referred to as HASH_SIZE.
+
+## Identifying Issuance Logs {#log-ids}
+
+A CA can operate one or more issuance logs. Each log that the CA operates is assigned a log number, which is a positive 64-bit integer, greater than zero. Log numbers must be unique per CA.
+
+The combination of the CA issuer ID and the log number uniquely identify an issuance log.
+
+TODO: Log numbers (as written) are positive instead of non-negative. 0 is intentionally excluded to force `serialNumbers` to have more than 64 bits and ensure that the parsing code for certificate consumers properly handles extracting the log number. This has a side effect of making all serial numbers be positive integers, which has another benefit described in the TODO in {{certificate-format}}.
```
davidben: Out-of-date TODO?

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1301,18 +1318,20 @@ When verifying the signature of an X.509 certificate (Step (a)(1) of {{Section 6
 
 1. Decode the `signatureValue` as an MTCProof, as described in {{certificate-format}}.
 
-1. Let `index` be the certificate's serial number. If `index` is contained in one of the relying party's revoked ranges ({{revocation-by-index}}), abort this process and fail verification.
+1. Let `index` be the lowest 64 bits of the certificate's serial number and let `log_number` be the serial number bit shifted right 64 bits.
```
davidben: Instead of "lowest", "least significant"? Although that gets into silly questions like:
* Is the type of the least significant bits list<bit> or integer?
* What happens if the serial number is negative, since 5280 says "Certificate users SHOULD be prepared to gracefully handle such certificates" and most stacks tolerate it.

What about:

```suggestion
1. Let `serial` be the certificate's serial number. If `serial` is negative, abort this process and fail verification.
1. Let `index` be `serial` modulo 2<sup>64</sup>, and let `log_number` be `serial >> 64`.
```

Should we reject `log_number` zero?

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1301,18 +1318,20 @@ When verifying the signature of an X.509 certificate (Step (a)(1) of {{Section 6
...
    1. Set `subjectPublicKeyInfoHash` to the hash of the DER encoding of `subjectPublicKeyInfo`.
 
 1. Construct a MerkleTreeCertEntry of type `tbs_cert_entry` with contents the TBSCertificateLogEntry. Let `entry_hash` be the hash of the entry, `MTH({entry}) = HASH(0x00 || entry)`, as defined in {{Section 2.1.1 of !RFC9162}}.
 
 1. Let `expected_subtree_hash` be the result of evaluating the MTCProof's `inclusion_proof` for entry `index`, with hash `entry_hash`, of the subtree described by the MTCProof's `start` and `end`, following the procedure in {{evaluating-a-subtree-inclusion-proof}}. If evaluation fails, abort this process and fail verification.
 
-1. If `[start, end)` matches a trusted subtree ({{trusted-subtrees}}), check that `expected_subtree_hash` is equal to the trusted subtree's hash. Return success if it matches and failure if it does not.
+1. If `[start, end)` matches a trusted subtree ({{trusted-subtrees}}) for the log identified by `issuer` and `log_number` ({{log-ids}}), check that `expected_subtree_hash` is equal to the trusted subtree's hash. Return success if it matches and failure if it does not.
```
davidben: It does feel odd to not say anything anywhere about what RPs and CAs are expected to do with this. We don't actually expect the RP to have subtrees configured for every one of the range of logs that exist. (Maybe we can leave that in a TODO where this is all set up? I dunno.)

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1966,7 +1987,7 @@ In the case when `fn` is `sn` in step 5, the condition in step 7.2.1 is always f
 
 ## Subtree Signed Note Format
 
-A subtree, with signatures, can be represented as a signed note {{SIGNED-NOTE}}. Trust anchor IDs can be converted into log origins and cosigner names by concatenating the ASCII string `oid/1.3.6.1.4.1.` and the ASCII representation of the trust anchor ID. For example, the checkpoint origin for a log named `32473.1` would be `oid/1.3.6.1.4.1.32473.1`.
+A subtree, with signatures, can be represented as a signed note {{SIGNED-NOTE}}. A log origin can be constructed from the log's CA issuer ID and log number ({{log-ids}}) in the following manner: The log origin is the concatenation of the ASCII string `oid/1.3.6.1.4.1.`, the CA issuer ID as its trust anchor ID's ASCII representation, the byte `0x2f` (ASCII character "/"), and the log number as an ASCII string in decimal with no leading zeroes.
```
davidben: While we _have_ gotten rid of the ASCII-to-binary conversion now, I think it is still better to just append the number as a dot. The spec as it exists right now does not technically admit a path component after the OID.
https://github.com/C2SP/C2SP/blob/main/tlog-cosignature.md#:~:text=For%20ecosystems%20that%20use%20OIDs%20for%20identification%2C%20the%20key%20name%20MAY%20be%20the%20string%20oid/%20followed%20by%20an%20OID%20in%20dotted%20decimal%20form.

(Also just seems unnecessary to invent a second kind of combiner because we _already_ have a way to append integers.)

draft-ietf-plants-merkle-tree-certs.md
davidben: Oh we also are missing a story for whether the client accepts that log number. The client is going to need a minimum log number (to close out the bad generations) and probably an optional maximum one.

davidben:
> The document doesn't specify the protocol between the witness and the CA, but it does say "A cosigner MAY expose a private interface for the CA, to reduce denial-of-service risk" and also mentions the idea of there being witnesses that the CA doesn't communicate with at all. I understand your point as being that it's possible to avoid this happening, but my point is that it's a very big footgun that's non-obvious and seems to be consistent with the text of the document.

Even if the cosigner is private to the CA, if you believe this massive-collateral-damage situation is real, you have to also believe that the CA wouldn't want this to happen. In that world, it is absolutely the case that the CA would be interested in pushing the witness forward.

This document also doesn't specify the protocol because it's a general mechanism, not "here is everything you need to know about this one PKI I'm interested in". Anything using the tlog ecosystem (including the Web PKI) will presumably use the tlog protocols. Those _are_ specified: https://c2sp.org/tlog-witness https://c2sp.org/tlog-mirror

(As it turns out, what that paragraph was actually talking about was the subtree signer oracle, as that doesn't have the forward-progress-always property that witnesses have. So this is really all a distraction. Maybe we should just delete that whole paragraph.)

But, regardless, this all seems to be unrelated to the topic at hand. Your overall claim seems to be that we should disregard witness-based PKIs, which is a bold claim that cuts out a large number of applications. (E.g. the interest from folks in the WG for applying some of this to private PKIs.) The argument for this claim is that there is a theoretical situation that might result in a massive collateral damage. However, that theoretical situation is, in itself, only plausible if everyone acts against their incentives to reduce risk. Moreover, even if that all went wrong, there is a clear technical mechanism to avoid the collateral damage: only close off that one witness's view of the broken log.

> My understanding is that the tlog-mirror doesn't produce a signature over a portion of the log until it has durably persisted all of the log entries. That's the only guarantee that we need for this to work. Support for a reconciliation operation can be added to tlog-mirror in a backwards-compatible way, and without affecting anything about how MTC works.

Yes, when tlog-mirror gets stuck, it has not _cryptographically_ committed to anything what would prevent it from resetting the pending checkpoint. That does not mean resetting state solves the problem. Unless it can be integrated into the design in a safe, automatic, and scaleable way, we're talking about a manual intervention, across multiple parties in the ecosystem. That will take a ton of coordination across the ecosystem, during which time there is a PKI outage. That's not operationally acceptable.

By contrast, the log generations approach maintains availability (the next few generations would be active already) while the ecosystem reacts to the incident. This matches how we handle PKIs today.

Now, I say all this, and I actually _do_ think making these protocols more robust, where possible, is a good idea. If we can make mirrors hard to wedge, awesome. If it takes a manual but still fairly safe "drop temporary state and kill all pending update requests" button somewhere, maybe that's still useful. But that is unchartered territory and added risk. The right place to start is aligned with what we have experience with.

> The idea that this can be removed later is only true from the perspective of an individual browser. Monitors need to care about what a long tail of browsers might accept. So once this is added and some swath of versions of browsers support it, monitors will need to support it for as long as any of those browser-versions have non-negligible usage. If you don't add this feature now, you won't have to slowly phase it out of the ecosystem later.

"Later" is much easier than you think. Today, every CT-enforcing client has an enforcement shut-off after 10 weeks. This is an operational necessity because we do not have a story today for clients with arbitrarily old log lists to continue to function.

I really dislike this mechanism, and am excited because MTCs finally gives a plausible path to not needing it. (See all my vague muttering about cosigner agility and negotiation. Also compare CT's need for year-sharded logs and an MTC cosigner's identity being stable.)

However, it is totally implausible, with the timelines we're working with, that we'll remove it in day 1 of MTCs. I expect the initial MTC story will be:
* Clients that do not require transparency cosigners at all (analog of today's non-CT-enforcing clients)
* Clients that require transparency cosigners, but disable this after being stale (analog of today's CT-enforcing clients)

That means later here is at most 10 weeks. Even as things improve here, think about the rest of the picture. The reason monitors care about monitoring all this stuff is that we want to find bad certs. But finding a bad cert doesn't automatically make it better. It's an alerting mechanism. The security enforcement comes later, with actions like...
* Maybe the cert is revoked and the revocation is pushed to clients
* Maybe a CA is found untrustworthy and a removal is pushed to clients

Transparency and monitoring cannot help stale clients. So we're already not talking about the long-tail on the time axis.

Bren2010:
> But, regardless, this all seems to be unrelated to the topic at hand. Your overall claim seems to be that we should disregard witness-based PKIs, which is a bold claim that cuts out a large number of applications.

I don't believe I claimed that. I'm pointing out a problem and proposing a solution. There are multiple solutions to the same problem though, and we can compare them.

You could add text to the document saying that all cosigners should always be kept relatively up to date. You could specify cosigner-based or subtree-based revocation. These are both direct, reasonable solutions to parts of what I stated. However, you'd want to do at least one of these regardless of whether you implement log generations or reconciliation, so these aren't points of differentiation.

The benefits of reconciliation over log generations are:
- Reconciliation requires no protocol mechanism in MTC and is easier to evolve over time.
- With log generations, monitoring any combination of (CA, log generation, mirror) could result in observing different certificates. With reconciliation, there are no log generations and it doesn't long-term matter which mirror you use, so it's much simpler to observe all certificates issued by a CA.

The downside of reconciliation over log generations is:
- It doesn't work for witnesses. However, I'd say that fits into a larger pattern of significant downsides that come from using witnesses. Another one being: the ability of a CA to cosign some certificate and then refuse to publish it. That's a much weaker transparency property than CT provides today. If witnesses are only meant for less robust, private PKI-type deployments, maybe it's fine that reconciliation isn't an option for them?

> Unless it can be integrated into the design in a safe, automatic, and scaleable way, we're talking about a manual intervention, across multiple parties in the ecosystem.

The reconciliation operation is essentially: "given list A and list B, verify that B is a superset of A." That can obviously be automated in a lot of ways, and generally with time complexity O(n log n) and space O(n). That's the same time and space complexity that running a mirror already takes, so I would generally expect this to be as scalable as the rest of the mirror operations. If you think it has a realistic chance of getting merged, I'm happy to write a proposal to c2sp.

draft-ietf-plants-merkle-tree-certs.md
nharper: Removed.

draft-ietf-plants-merkle-tree-certs.md
nharper: Removed `logNumber`.

draft-ietf-plants-merkle-tree-certs.md
nharper: Removed the TODO.

draft-ietf-plants-merkle-tree-certs.md
nharper: Changed "lowest" to "least significant". I prefer "least significant" over "modulo 2^64", because we already least-significant bits in the draft, but we don't have modulo operators. Also added language to reject log number 0.

Added a bit to {{trust-anchors}} that a client can have a minimum log number.

draft-ietf-plants-merkle-tree-certs.md
nharper: I don't understand this comment. Subtrees are an optional optimization — it's fine for `[start, end)` to not match any trusted subtrees (because the RP has none at all, or the RP has none for that log), and we go on to the next step to verify the standalone cert.

draft-ietf-plants-merkle-tree-certs.md
nharper: I'm not a fan of "issuance log" referring to all log numbers together made up of a bunch of trees, because that diverges from what we consider a "log" to be today in CT (RFC 6962 and Static CT) and TLOG.

I added a TODO to consider renaming "issuer ID", because I agree that this term can be confusing.

draft-ietf-plants-merkle-tree-certs.md
nharper: Switched it to dot, and added a TODO to explicitly document that we're using a full OID arc (though since the last component here is always a log number and this constructed OID isn't used anywhere else, we probably don't need to take over the whole OID arc).

draft-ietf-plants-merkle-tree-certs.md
davidben: Ah yeah, with https://github.com/ietf-plants-wg/merkle-tree-certs/pull/220, I now especially agree that calling the top-level thing a "log ID" makes no sense.

Honestly, after that PR, now all we have is:
* The CA has a trust anchor ID
* This implicitly gives logs and landmarks IDs

So maybe we just say "The CA has a trust anchor ID" and "the CA's ID {{ca-ids}}"?

draft-ietf-plants-merkle-tree-certs.md
davidben: I think I was mostly thinking that {trusted-subtrees} maybe should say something about how you're not expected to ship subtrees for every log, but I guess that all falls under "this PR doesn't really say when you use log numbers".

davidben (commented):
Suspect this will need a bit of a rebase with the CA format added in.

nharper:
Rebased; "CA issuer ID" is no more (it's now "CA ID").

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -893,46 +893,50 @@ A snapshot of the log is known as a *checkpoint*. A checkpoint is identified by
...
 Cosigners ({{cosigners}}) sign assertions about the state of the issuance log. A Merkle Tree CA operates a combination of an issuance log and CA cosigner ({{certification-authority-cosigners}}) that authenticates the log state and certifies the contents. External cosigners may also be deployed to assert correct log operation or provide other services to relying parties ({{trusted-cosigners}}).
 
-## Log Parameters
-
-An issuance log has the following parameters:
-
-* A log ID, which uniquely identifies the log. See {{log-ids}}.
-* A collision-resistant cryptographic hash function. SHA-256 {{!SHS=DOI.10.6028/NIST.FIPS.180-4}} is RECOMMENDED.
-* A minimum index, which is the index of the first log entry which is available. See {{log-pruning}}. This value changes over the lifetime of the log.
+## Relationship with Certification Authorities {#ca-ids}
```
davidben, line 896: This framing reads oddly to me. I suspect we actually want something like a Certification Authorities section that says "these are the parts of a CA" and then maybe Issuance Logs is a subsection of that. Maybe moving Representing Certification Authorities too. Dunno how we want to sequence doing that surgery, #227, and sending an email to the list. (I'm inclined to do that last one sooner than later at this point.)

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -893,46 +893,50 @@ A snapshot of the log is known as a *checkpoint*. A checkpoint is identified by
...
 
-Throughout this document, the hash algorithm in use is referred to as HASH, and the size of its output in bytes is referred to as HASH_SIZE.
+Each issuance log is run by a Certification Authority. A CA can run multiple related issuance logs. Each log is assigned a positive integer to identify it, and these log numbers MUST be assigned in increasing order.
 
-## Log IDs
+Each CA that operates one or more issuance log has a *CA ID* to identify it. This CA ID is a trust anchor ID {{!I-D.ietf-tls-trust-anchor-ids}}, and its entire OID arc is allocated for the issuance logs run by that CA.
```
davidben: s/one or more issuance log/one or more issuance logs/

Though "Each CA that issues Merkle Tree Certificates" or "Each Merkle Tree Certificate CA" might get more to the point.

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -893,46 +893,50 @@ A snapshot of the log is known as a *checkpoint*. A checkpoint is identified by
...
 
+## Log Parameters
+
+An issuance log has the following parameters:
+
+* A log ID, constructed from the CA ID and log number, which uniquely identifies the log. See {{ca-ids}}.
+* A collision-resistant cryptographic hash function. SHA-256 {{!SHS=DOI.10.6028/NIST.FIPS.180-4}} is RECOMMENDED.
+* A minimum index, which is the index of the first log entry which is available. See {{log-pruning}}. This value changes over the lifetime of the log.
+
+Throughout this document, the hash algorithm in use is referred to as HASH, and the size of its output in bytes is referred to as HASH_SIZE.
```
davidben, line 948: The hash really is a parameter of the whole CA. We don't let them switch hashes across their logs (the CA format will go haywire).

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1316,15 +1320,15 @@ Proof sizes grow logarithmically, so 32 hashes, or 1024 bytes, is sufficient for
...
 
-* The `subject` field MUST be the CA issuance log's log ID as a PKIX distinguished name, as described in {{log-ids}}.
+* The `subject` field MUST be the CA ID as a PKIX distinguished name, as described in {{ca-ids}}.
 
 * The `subjectPublicKeyInfo` field MUST be the public key of the CA cosigner {{certification-authority-cosigners}}.
 
 * The `extensions` field MUST contain a critical extension of type id-pe-mtcCertificationAuthority, defined below.
 
-* The subject key identifier extension ({{Section 4.2.1.2 of !RFC5280}}), if present, SHOULD be set to CA's log ID {{log-ids}}. The log ID is encoded in its binary representation, as defined in {{Section 3 of !I-D.ietf-tls-trust-anchor-ids}}.
+* The subject key identifier extension ({{Section 4.2.1.2 of !RFC5280}}), if present, SHOULD be set to the CA ID {{ca-ids}}. The OID is encoded in its binary representation, as defined in {{Section 3 of !I-D.ietf-tls-trust-anchor-ids}}.
```
```suggestion
* The subject key identifier extension ({{Section 4.2.1.2 of !RFC5280}}), if present, SHOULD be set to the CA ID {{ca-ids}}. The CA ID is encoded in its binary representation, as defined in {{Section 3 of !I-D.ietf-tls-trust-anchor-ids}}.
```

(That section doesn't define a binary representation of an OID, just a trust anchor ID.)

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1379,11 +1383,17 @@ This section discusses how relying parties verify Merkle Tree Certificates.
...
 
-* The log ID ({{log-ids}})
-* The log hash algorithm, e.g. SHA-256
+* The CA's ID ({{ca-ids}})
+* The hash algorithm used in each log, e.g. SHA-256
 * The CA cosigner, and any other supported cosigners, as pairs of cosigner ID and public key
 * A policy on which combinations of cosigners to accept in a certificate ({{trusted-cosigners}})
-* An optional list of trusted subtrees, with their hashes, that are known to be consistent with the relying party's cosigner requirements ({{trusted-subtrees}})
+
+A relying party MAY be configured on a per-CA basis with a minimum log number, such that it will only trust certificates from logs with a log number greater than or equal to that minimum.
```
davidben, line 1400: Do we also allow them to be configured with maximum log numbers?

Edit: Ooh, or see the idea about revoked serial ranges instead of revoked index ranges below.

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1406,7 +1416,16 @@ When verifying the signature of an X.509 certificate (Step (a)(1) of {{Section 6
 
 1. Decode the `signatureValue` as an MTCProof, as described in {{certificate-format}}.
 
-1. Let `index` be the certificate's serial number. If `index` is contained in one of the relying party's revoked ranges ({{revocation-by-index}}), abort this process and fail verification.
+1. Let `serial` be the certificate's serial number. If `serial` is negative, abort this process and fail verification.
+
+1. Let `index` be the least significant 64 bits of `serial` and let `log_number` be `serial >> 64`.
+
+1. Let `logID` be the log ID constructed from the CA ID in `issuer` and the `log_number` ({{ca-ids}})
```
davidben: Nit: A little odd to use `logID` but then `log_number`. (Underscores vs camelCase.)

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1406,7 +1416,16 @@ When verifying the signature of an X.509 certificate (Step (a)(1) of {{Section 6
...
+
+1. If any of the following conditions are true, abort this process and fail verification:
+   * The `log_number` is zero
+   * The `log_number` is smaller than the minimum log number for that CA ({{trust-anchors}})
+   * `index` is contained in one of the relying party's revoked ranges for `logID` ({{revocation-by-index}})
```
davidben: A thought: we said `minSerial` for the CA format to kinda get around having a two-level index now... should we do the same here too? That avoids having to deal with this, or answer annoying questions like "what if I want to cut off all log numbers past a certain point, that's an infinite number of ranges".

(Cutting off all log numbers past a certain point needed to do SCTNotAfter.)

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1379,11 +1383,17 @@ This section discusses how relying parties verify Merkle Tree Certificates.
...
-* The log hash algorithm, e.g. SHA-256
+* The CA's ID ({{ca-ids}})
+* The hash algorithm used in each log, e.g. SHA-256
 * The CA cosigner, and any other supported cosigners, as pairs of cosigner ID and public key
 * A policy on which combinations of cosigners to accept in a certificate ({{trusted-cosigners}})
-* An optional list of trusted subtrees, with their hashes, that are known to be consistent with the relying party's cosigner requirements ({{trusted-subtrees}})
+
+A relying party MAY be configured on a per-CA basis with a minimum log number, such that it will only trust certificates from logs with a log number greater than or equal to that minimum.
+
+Additionally, a relying party MAY be configured with the following for each log operated by the CA:
```
davidben: For *each* log seems a little wonky. There are infinitely many potential logs, and I can't configure an infinite number of lists. (Even if I have non-zero subtrees for one log, you could read this as saying I need an infinite number of empty `vector<TrustedSubtree>`s.)

What if we said, outside of any "for each log" bits, that you might have a list of trusted subtrees, and that each trusted subtree told you:
* Log number
* `[start, end)` coordinates
* The hash

(Come to think of it, I don't know if we ever said what was in a single trusted subtree. Whoops.)

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1427,7 +1446,8 @@ In this procedure, `entry_hash` can equivalently be computed in a single pass fr
 
 1. Initialize a hash instance.
 1. Write the big-endian, two-byte `tbs_cert_entry` value to the hash.
-1. Write the TBSCertificate contents octets to the hash, up to the `subjectPublicKeyInfo` field.
+1. Write the TBSCertificate's `version` field to the hash.
+1. Write the TBSCertificate contents octets to the hash, startin from the `issuer` field, up to (but not including) the `subjectPublicKeyInfo` field.
```
davidben: s/startin/starting/ but I think we can actually revert these back to "1. Write the TBSCertificate contents octets to the hash, up to the `subjectPublicKeyInfo` field." now that the log number is out of the entry.

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1487,15 +1507,15 @@ The relying party SHOULD incorporate its trusted subtree configuration in applic
...
+When a relying party is first configured to trust an issuance log, it SHOULD be configured to revoke all entries from zero up to but not including the first available unexpired certificate at the time. This revocation SHOULD be periodically updated as entries expire and logs are pruned ({{log-pruning}}). In particular, when CAs prune entries, relying parties SHOULD be updated to revoke all newly unavailable entries. This gives assurance that, even if some unavailable entry had not yet expired, the relying party will not trust it. It also allows monitors to start monitoring a log without processing expired entries.
```
davidben, line 1519: The first sentence is a little goofy because relying parties are configured to trust CAs, not issuance logs. (The above comment about setting the minimum index might help.)

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1487,15 +1507,15 @@ The relying party SHOULD incorporate its trusted subtree configuration in applic
...
 
 When a CA is found to be untrustworthy, relying parties SHOULD remove trust in that CA. To minimize the compatibility impact of this mitigation, index-based revocation can be used to only distrust entries after some index, while leaving existing entries accepted. This is analogous to the {{SCTNotAfter}} mechanism used in some PKIs.
 
-The revocation mechanism in this section is complementary to certificate-level revocation mechanisms. Because Merkle Tree certificates use log indices as serial numbers, existing revocation mechanisms like CRLs {{!RFC5280}} and OCSP {{!RFC6960}} apply unchanged.
+The revocation mechanism in this section is complementary to certificate-level revocation mechanisms. Because Merkle Tree certificates are uniquely identified by their serial number and issuer, existing revocation mechanisms like CRLs {{!RFC5280}} and OCSP {{!RFC6960}} apply unchanged.
```
```suggestion
The revocation mechanism in this section is complementary to certificate-level revocation mechanisms. Because log entries are uniquely identified by their serial number and issuer, existing revocation mechanisms like CRLs {{!RFC5280}} and OCSP {{!RFC6960}} apply unchanged.
```

(Arguably the _certificates_ are not uniquely identified because you can have multiple forms of the same certificate.)

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -2104,7 +2124,7 @@ In the case when `fn` is `sn` in step 5, the condition in step 7.2.1 is always f
 
 ## Subtree Signed Note Format
 
-A subtree, with signatures, can be represented as a signed note {{SIGNED-NOTE}}. Trust anchor IDs can be converted into log origins and cosigner names by concatenating the ASCII string `oid/1.3.6.1.4.1.` and the ASCII representation of the trust anchor ID. For example, the checkpoint origin for a log named `32473.1` would be `oid/1.3.6.1.4.1.32473.1`.
+A subtree, with signatures, can be represented as a signed note {{SIGNED-NOTE}}. Trust anchor IDs can be converted into log origins and cosigner names by concatenating the ASCII string `oid/1.3.6.1.4.1.` and the ASCII representation of the trust anchor ID. For example, the log ID from a CA with ID `32473.1` and log number 8 is `32473.1.0.8` and its checkpoint origin is `oid/1.3.6.1.4.1.32473.1.0.8`.
```
davidben, line 2124: Hopefully we can just delete this section soon via #225

draft-ietf-plants-merkle-tree-certs.md
nharper: Yes, we really need a CA section, but let's punt that for a later PR.

draft-ietf-plants-merkle-tree-certs.md
nharper: Reworded to "Each Merkle Tree Certificate CA"

draft-ietf-plants-merkle-tree-certs.md
nharper: Done.

draft-ietf-plants-merkle-tree-certs.md
nharper: Switched to `log_id`

draft-ietf-plants-merkle-tree-certs.md
nharper: My intention wasn't that if the relying party chooses to do the MAY that they must do it for every log, but that they might want to be configured with information for some logs. I changed the wording from "each" to "per-log".

draft-ietf-plants-merkle-tree-certs.md
nharper: Reverted.

draft-ietf-plants-merkle-tree-certs.md
nharper: If the relying party is configured to only trust CAs (i.e. the CA ID and the key, and nothing else), then this doesn't matter. For this to apply, the relying party has to have trusted information about particular logs, hence it's configured to trust an issuance log (as part of its configuration for trusting a CA). I don't think this needs to change as part of this PR.

draft-ietf-plants-merkle-tree-certs.md
nharper: Fixed.

draft-ietf-plants-merkle-tree-certs.md
nharper: This looks like a candidate for moving to the CA section to be written in a future PR. In the meantime, I added a MUST that the same hash function be used for all logs under the same CA ID.

draft-ietf-plants-merkle-tree-certs.md
nharper: A relying party MAY choose to do all sorts of things not specified in the draft. I don't think we should specify configuring a maximum log number. The purpose of log numbers is to allow (stale) relying parties to trust arbitrarily large log numbers, and we don't want a stale config that has a maximum set for no good reason.

draft-ietf-plants-merkle-tree-certs.md
nharper: Added a TODO.

PR #215, opened 2026-04-23
Align signature scheme with tlog-cosignature

davidben:
This makes minor changes to the signature scheme to make it compatible with https://c2sp.org/tlog-cosignature, with the aim that MTCs can be more easily deployed by using unmodified tlog-based software.

This does not add a normative dependency on tlog, though we do non-normatively cite a tlog document so readers know the signature construction was chosen to be compatible. It also remains unnecessary to use tlog to implement MTCs. This is just defining the format for the signature input.

While I'm here, some of the discussion about cosignatures was... very confusing. We were talking about keeping a current checkpoint and everything, without defining cosigner roles. Let's leave exact state to the cosigner roles and instead, here, we can just abstractly define what you need to guarantee.

"Cosigners SHOULD publish their current checkpoint, along with the checkpoint signature" was particularly weird since it doesn't say how to publish it. While it is true for basically all cosigners in CT-like deployments, it should be for the individual cosigner roles to define.

### DIFF
--- a/draft-ietf-plants-merkle-tree-certs.md
+++ b/draft-ietf-plants-merkle-tree-certs.md
@@ -1010,7 +1010,7 @@ This signature format is designed to be compatible with the ML-DSA-44 signature 
 
 `label` is a fixed prefix for domain separation. Its value MUST be the string `subtree/v1`, followed by a newline (U+000A), followed by a zero byte (U+0000).
 
-`cosigner_name` and `log_origin`, are computed from the cosigner ID and the issuance log's ID ({{log-ids}}), respectively. The contain the concatenation of:
+`cosigner_name` and `log_origin` are computed from the cosigner ID and the issuance log's ID ({{log-ids}}), respectively. They contain the concatenation of:
 
 * The 16-byte ASCII string `oid/1.3.6.1.4.1.`
 * The trust anchor ID's ASCII representation ({{Section 3 of !I-D.ietf-tls-trust-anchor-ids}})
@@ -1338,7 +1338,7 @@ When verifying the signature of an X.509 certificate (Step (a)(1) of {{Section 6
 
 1. If `[start, end)` matches a trusted subtree ({{trusted-subtrees}}), check that `expected_subtree_hash` is equal to the trusted subtree's hash. Return success if it matches and failure if it does not.
 
-1. Otherwise, check that the MTCProof's `signatures` contain a sufficient set of valid signatures from cosigners to satisfy the relying party's cosigner requirements ({{trusted-cosigners}}). Unrecognized cosigners MUST be ignored. Signatures are verified as described in {{signature-format}}. The `subtree_hash` field of the CosignedMessage is set to `expected_subtree_hash`. The `timestamp` field is set to zero.
+1. Otherwise, check that the MTCProof's `signatures` contain a sufficient set of valid signatures from cosigners to satisfy the relying party's cosigner requirements ({{trusted-cosigners}}). Unrecognized cosigners MUST be ignored. Signatures are verified as described in {{signature-format}}. Reconstruct the CosignedMessage from MTCProof's `start` and `end`, the cosigner ID for `cosigner_name`, the log ID for `log_origin`, `expected_subtree_hash` for `subtree_hash`, and `timestamp` set to zero`.
 
 This procedure only replaces the signature verification portion of X.509 path validation. The relying party MUST continue to perform other checks, such as checking expiry.
 
@@ -2004,7 +2004,7 @@ Each note signature has a key name of the cosigner name. The signature's key ID 
 key ID = SHA-256(key name || 0x0A || 0xFF || "mtc-subtree/v1")[:4]
 ~~~
 
-[[TODO: When all this to C2SP, allocate a codepoint for a more idiomatic key ID construction.]]
+[[TODO: When all this moves to C2SP, allocate a codepoint for a more idiomatic key ID construction.]]
 
 ## Requesting Subtree Signatures
 
--- a/draft-ietf-plants-merkle-tree-certs.md
+++ b/draft-ietf-plants-merkle-tree-certs.md
@@ -1022,11 +1022,16 @@ This is equivalent to the concatenation of:
 
 For example, the trust anchor ID 32473.1 would be encoded as the ASCII string `oid/1.3.6.1.4.1.32473.1`.
 
-`start` and `end` MUST define a valid subtree of the log, and `subtree_hash` MUST be the subtree's hash value in the cosigner's view of the log. If `start` is non-zero, then `timestamp` MUST be zero.
+`start` and `end` MUST define a valid subtree of the log, and `subtree_hash` MUST be the subtree's hash value in the cosigner's view of the log.
 
-If `timestamp` is non-zero, it MUST be the time that the signature was produced, represented as seconds since the Epoch as defined in Section 4.19 of Volume 1 of {{!POSIX=DOI.10.1109/IEEESTD.2024.10555529}}. Additionally `end` MUST be the size of the largest consistent tree that the cosigner has observed for the log.
+If `timestamp` is non-zero, it MUST be the time that the signature was produced. This time is represented as seconds since the Epoch, as defined in Section 4.19 of Volume 1 of {{!POSIX=DOI.10.1109/IEEESTD.2024.10555529}}. Additionally, if `timestamp` is non-zero, the following MUST be true:
 
-If `timestamp` is zero, no statement is made about the signing time or largest observed tree.
+* `start` MUST be zero.
+* `end` MUST be the size of the largest consistent tree that the cosigner has observed for the log.
+
+`timestamp` MAY be zero, in which case no additional constraints are placed on `start` or `end`, and no statement is made about the signing time or largest observed tree.
+
+### Signature Semantics
 
 Before signing a subtree of some log, the cosigner MUST ensure that `subtree_hash` is consistent with its view of the log. Different cosigner roles may obtain this assurance differently. For example:
 
@@ -1038,8 +1043,7 @@ In both cases, the cosigner MUST ensure that, as it updates its view of the log,
 
 When a cosigner signs a subtree, it is held separately responsible *both* for the subtree being consistent with its other signatures, *and* for the cosigner-specific additional statements. That is, if a cosigner signs an inconsistent subtree, it is held responsible for its additional statements on all entries in the inconsistent subtree, even if some other signed subtree exists that asserts different entries.
 
-Subtree signatures can be used to sign timestamped log checkpoints by setting `start` to zero and a non-zero `timestamp`. These signatures are not directly used in Merkle Tree Certificates ({{certificate-format}}), but cosigners MAY generate them, subject to the rules above, as part of other functions in a PKI. This may include log serving or integrating an issuance log into a transparency ecosystem. For example, {{TLOG-TILES}} and {{TLOG-WITNESS}} use such signatures.
-
+Subtree signatures can be used to sign timestamped log checkpoints with a non-zero `timestamp`. A signature with a non-zero `timestamp` asserts the complete state of the cosigner's view of the log at a given time. These signatures are not directly used in Merkle Tree Certificates ({{certificate-format}}), but cosigners MAY generate them, subject to the rules above, as part of other functions in a PKI. This may include log serving or integrating an issuance log into a transparency ecosystem. For example, {{TLOG-TILES}} and {{TLOG-WITNESS}} use such signatures.
 
 ### Signature Algorithms
 
--- a/draft-ietf-plants-merkle-tree-certs.md
+++ b/draft-ietf-plants-merkle-tree-certs.md
@@ -984,13 +984,13 @@ This section defines a log *cosigner*. A cosigner follows some append-only view 
 
 A cosignature MAY implicitly make additional statements about a subtree, determined by the cosigner's role. This document defines one concrete cosigner role, a CA cosigner ({{certification-authority-cosigners}}), to authenticate the log and certify entries. Other documents and specific deployments may define other cosigner roles, to perform different functions in a PKI. For example, {{TLOG-WITNESS}} defines a cosigner that only checks the log is append-only, and {{TLOG-MIRROR}} defines a cosigner that mirrors a log.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1980,13 +2003,7 @@ Each note signature has a key name of the cosigner name. The signature's key ID
 key ID = SHA-256(key name || 0x0A || 0xFF || "mtc-subtree/v1")[:4]
```
FiloSottile, line 2008:
```suggestion
key ID = SHA-256(key name || 0x0A || 0xFF || "subtree/v1")[:4]
```

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -983,39 +990,53 @@ A single cosigner, with a single cosigner ID and public key, MAY generate cosign
...
+`timestamp`, if non-zero, indicates both that the signature was produced at the time, and that `end` is the size of largest consistent tree the cosigner has observed for the log. If `timestamp` is zero, no statement is made about signing time or the largest such tree. If `start` is non-zero, `timestamp` MUST be zero.
 
-Cosigners SHOULD publish their current checkpoint, along with the checkpoint signature.
+The resulting signature is known as a *subtree signature*. When `start` is zero, the resulting signature describes the checkpoint with tree size `end` and is also known as a *checkpoint signature*. All subtree signatures used in this document's certificate construction ({{certificate-format}}) have a `timestamp` of zero.
```
aarongable: These two paragraphs make it clear that `start` and `timestamp` are mutually exclusive, but it's (to me) very unclear _why_ that is the case. This is especially confusing because the second paragraph assigns a name ("checkpoint signature") to one of these kinds of certs (where `start` is zero), and then says that this document only uses the _other_ kind of signature (where `timestamp` is zero), which doesn't have a name. Why assign a name to one thing, and then say that we're using the other thing?

davidben: > These two paragraphs make it clear that start and timestamp are mutually exclusive, but it's (to me) very unclear why that is the case.

I mean the narrow answer for "why that is the case" is "because the proposed spec text defines it that way", but that is unhelpful. 🙃

The answer for "why is this defined that way" is that there are two incompatible definitions for what the timestamp could mean:

1. I generated this signature at time T, so my view of the tree was *at least as big as* [start, end) at the time T.
2. I generated this signature at time T over *my entire view of the tree*. Not only was the tree *at least as big* as [start, end) at time T, but *end is the size of the tree as of T*.

tlog-witness uses the second semantics for checkpoint signatures. (There are no subtree signatures in tlog yet, we need to go back and add them.) This is useful for monitoring because you know that you've received a fresh status from the witness, rather than a replay of an old one.

However, the second semantics are kinda incompatible with subtree signatures embedded in certificates for a couple reasons:

First, say you go from tree size 1023 to tree size 1100 in a batch of certs. You ideally construct that cert by signing [1023, 1100), but that range is not a subtree. If you were to build a Merkle Tree over it, you could not efficiently prove it consistent with [0, 1100). So we have this construction where it's split into a [pair of subtrees](https://ietf-plants-wg.github.io/merkle-tree-certs/draft-ietf-plants-merkle-tree-certs.html#name-arbitrary-intervals), specifically [1023, 1024) and [1024, 1100). But that first subtree is necessarily behind the witness position, so it cannot possibly meet condition 2. I would needed to have gone back in time to find when my view of the tree was at 1024.

Even without this two-subtrees construction, getting certificate-embedded signatures to meet condition 2 seems very tricky. Witnesses only move forward. Say you're trying to mint certificates anchored on [1023, 1100), so you catch the witnesses up to size 1100. If some witness was _already_ past 1100, you cannot rewind it. It will never again be willing to claim end = 1100 under condition 2.

Now, if "you" are the CA, okay, presumably you didn't add anything past 1100, so the witness cannot possibly be past 1100. But this assumes that signature collection happens in lockstep with appending to the tree. Maybe your issuing thread has gone ahead and moved the log forward. Or maybe "you" isn't the CA at all and this is a site operator trying to collect an extra cosignature that the CA missed.

It seems operationally a lot simpler to say:

* First you catch the witness up to *anywhere* past the subtree. If it was already caught up, or somehow got ahead, that's cool. It just needs to be contained in it.

* The witness exposes some [sign subtree](https://ietf-plants-wg.github.io/merkle-tree-certs/draft-ietf-plants-merkle-tree-certs.html#name-requesting-subtree-signatur) endpoint to request subtree signatures for anything contained in the current position.

Those signatures cannot meet condition (2), so they must be distinguishable from checkpoint signatures. They _could_ in theory meet condition (1), but there didn't seem to be a clear use case for condition (1), so the signature scheme just says "your options are to assert condition (2) or to make no assertion". 🤷 

>  This is especially confusing because the second paragraph assigns a name ("checkpoint signature") to one of these kinds of certs (where start is zero), and then says that this document only uses the other kind of signature (where timestamp is zero), which doesn't have a name. Why assign a name to one thing, and then say that we're using the other thing?

It's an awkward artifact of the scoping. This document defines certificates, which only use subtree signatures. But we expect it will often (certainly in the Web PKI) be used as part of a broader system (log mirroring and witnessing) that _does_ generate checkpoint signatures.

If we just say "timestamp is zero" and move on, then it's a little unclear what it means if someone signs something where timestamp is non-zero. Does it mean:

* This is an illegal use of the CA's key and therefore misissuance? That would be bad. We _want_ the CA to generate checkpoint signatures.
* This is out of scope for definition of the CA and therefore we don't care what the signature contains? That would be bad. If the CA signs a checkpoint containing misissued stuff, that stuff still in scope for misissuance.

Given that, and the desire to avoid a normative reference to tlog-cosignature, I figured I'd just redefine it here. Though the wording is awkward so I'll take another pass at it.

davidben: Took another pass at it. WDYT?

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -977,45 +984,62 @@ This section defines a log *cosigner*. A cosigner follows some append-only view
...
+    HashValue subtree_hash;
+} CosignedMessage;
 ~~~
 
-`log_id` MUST be the issuance log's ID ({{log-ids}}), in its binary representation ({{Section 3 of !I-D.ietf-tls-trust-anchor-ids}}). `start` and `end` MUST define a valid subtree of the log, and `hash` MUST be the subtree's hash value in the cosigner's view of the log. The `label` is a fixed prefix for domain separation. Its value MUST be the string `mtc-subtree/v1`, followed by a newline (U+000A), followed by a zero byte (U+0000). `cosigner_id` MUST be the cosigner ID, in its binary representation.
+This signature format is designed to be compatible with the ML-DSA-44 signature construction in {{TLOG-COSIGNATURE}}, but it supports signature algorithms other than ML-DSA-44 and tree hashes other than SHA-256.
+
+`label` is a fixed prefix for domain separation. Its value MUST be the string `subtree/v1`, followed by a newline (U+000A), followed by a zero byte (U+0000).
+
+`cosigner_name` and `log_origin`, are computed from the cosigner ID and the issuance log's ID ({{log-ids}}), respectively. The contain the concatenation of:
```
```suggestion
`cosigner_name` and `log_origin` are computed from the cosigner ID and the issuance log's ID ({{log-ids}}), respectively. They contain the concatenation of:
```

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -977,45 +984,62 @@ This section defines a log *cosigner*. A cosigner follows some append-only view
...
+This is equivalent to the concatenation of:
+
+* The four-byte ASCII string `oid/`
+* The trust anchor ID as a full OID, in dotted decimal notation
+
+For example, the trust anchor ID 32473.1 would be encoded as the ASCII string `oid/1.3.6.1.4.1.32473.1`.
+
+`start` and `end` MUST define a valid subtree of the log, and `subtree_hash` MUST be the subtree's hash value in the cosigner's view of the log. If `start` is non-zero, then `timestamp` MUST be zero.
+
+If `timestamp` is non-zero, it MUST be the time that the signature was produced, represented as seconds since the Epoch as defined in Section 4.19 of Volume 1 of {{!POSIX=DOI.10.1109/IEEESTD.2024.10555529}}. Additionally `end` MUST be the size of the largest consistent tree that the cosigner has observed for the log.
```
lukevalenta: Possible rephrasing to first state the wire-format rules, and then give the semantics:
```suggestion
If `start` is non-zero, then `timestamp` MUST be zero. Otherwise, `timestamp` MAY be zero or it MAY be the time that the signature was produced, represented as seconds since the Epoch as defined in Section 4.19 of Volume 1 of {{!POSIX=DOI.10.1109/IEEESTD.2024.10555529}}; in the latter case `end` MUST be the size of the largest consistent tree that the cosigner has observed for the log.

A non-zero `timestamp` is an assertion about the signing time and the largest observed tree; a zero `timestamp` makes no such statement.
```

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -977,45 +984,62 @@ This section defines a log *cosigner*. A cosigner follows some append-only view
...
+* The trust anchor ID's ASCII representation ({{Section 3 of !I-D.ietf-tls-trust-anchor-ids}})
+
+This is equivalent to the concatenation of:
+
+* The four-byte ASCII string `oid/`
+* The trust anchor ID as a full OID, in dotted decimal notation
+
+For example, the trust anchor ID 32473.1 would be encoded as the ASCII string `oid/1.3.6.1.4.1.32473.1`.
+
+`start` and `end` MUST define a valid subtree of the log, and `subtree_hash` MUST be the subtree's hash value in the cosigner's view of the log. If `start` is non-zero, then `timestamp` MUST be zero.
```
lukevalenta: If the edit below is accepted:
```suggestion
`start` and `end` MUST define a valid subtree of the log, and `subtree_hash` MUST be the subtree's hash value in the cosigner's view of the log.
```

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -977,45 +984,62 @@ This section defines a log *cosigner*. A cosigner follows some append-only view
...
+
+If `timestamp` is non-zero, it MUST be the time that the signature was produced, represented as seconds since the Epoch as defined in Section 4.19 of Volume 1 of {{!POSIX=DOI.10.1109/IEEESTD.2024.10555529}}. Additionally `end` MUST be the size of the largest consistent tree that the cosigner has observed for the log.
 
-The resulting signature is known as a *subtree signature*. When `start` is zero, the resulting signature describes the checkpoint with tree size `end` and is also known as a *checkpoint signature*.
+If `timestamp` is zero, no statement is made about the signing time or largest observed tree.
```
lukevalenta: If the edit above is accepted:
```suggestion
```

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1980,13 +2004,7 @@ Each note signature has a key name of the cosigner name. The signature's key ID
...
 ~~~
 
-A subtree whose `start` is zero can also be represented as a checkpoint {{TLOG-CHECKPOINT}}. A corresponding subtree signature can be represented as a note signature using a key ID computed as follows:
-
-~~~pseudocode
-key ID = SHA-256(key name || 0x0A || 0xFF || "mtc-checkpoint/v1")[:4]
-~~~
-
-The only difference between the two forms is the implicit transformation from the signed note text to the MTCSubtree structure.
+[[TODO: When all this to C2SP, allocate a codepoint for a more idiomatic key ID construction.]]
```
```suggestion
[[TODO: When all this moves to C2SP, allocate a codepoint for a more idiomatic key ID construction.]]
```

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1314,7 +1338,7 @@ When verifying the signature of an X.509 certificate (Step (a)(1) of {{Section 6
 
 1. If `[start, end)` matches a trusted subtree ({{trusted-subtrees}}), check that `expected_subtree_hash` is equal to the trusted subtree's hash. Return success if it matches and failure if it does not.
 
-1. Otherwise, check that the MTCProof's `signatures` contain a sufficient set of valid signatures from cosigners to satisfy the relying party's cosigner requirements ({{trusted-cosigners}}). Unrecognized cosigners MUST be ignored. Signatures are verified as described in {{signature-format}}. The `hash` field of the MTCSubtree is set to `expected_subtree_hash`.
+1. Otherwise, check that the MTCProof's `signatures` contain a sufficient set of valid signatures from cosigners to satisfy the relying party's cosigner requirements ({{trusted-cosigners}}). Unrecognized cosigners MUST be ignored. Signatures are verified as described in {{signature-format}}. The `subtree_hash` field of the CosignedMessage is set to `expected_subtree_hash`. The `timestamp` field is set to zero.
```
lukevalenta: Just to explicitly list everything that goes into CosignedMessage:
```suggestion
1. Otherwise, check that the MTCProof's `signatures` contain a sufficient set of valid signatures from cosigners to satisfy the relying party's cosigner requirements ({{trusted-cosigners}}). Unrecognized cosigners MUST be ignored. Signatures are verified as described in {{signature-format}}. Reconstruct the CosignedMessage from MTCProof's `start` and `end`, the cosigner ID for `cosigner_name`, the log ID for `log_origin`, `expected_subtree_hash` for `subtree_hash`, and `timestamp` set to zero`.
```

lukevalenta (approved):

draft-ietf-plants-merkle-tree-certs.md
davidben: Hmm. I feel like this makes it a lot harder to tell what a non-zero timestamp *means*. I'll take another pass at this and see if it works better.

lukevalenta: Eh yeah feel free to ignore!

davidben: How's this version?

lukevalenta: Reads well to me!

draft-ietf-plants-merkle-tree-certs.md
davidben: (Talked with Aaron out of band and it sounds like this is resolved.)

FiloSottile (approved):
💯

draft-ietf-plants-merkle-tree-certs.md
davidben: Oh this'll just get deleted with https://github.com/C2SP/C2SP/pull/245 anyway. This was back when I thought we needed a signed-note representation for subtrees and did some ad-hoc MTC-specific thing for the key IDs. I think, while it's (very, very temporarily) MTC-specific, having the "mtc-" prefix there makes sense.

(But mostly I figure I'll leave it alone since it's doomed to be removed.)

Issue #214, opened 2026-04-22
Make sure OCSP and CRLs can *actually* be applied as-is


davidben:
See Ilari's comments here: https://mailarchive.ietf.org/arch/msg/plants/DkZDqqE2r7qokdwd7bHg8Fg7NBY/

davidben:
Specifically reevaluate this after #96 is closed and we have an X.509 encoding for a CA.

sgmenda:
see also https://github.com/ietf-plants-wg/merkle-tree-certs/pull/210#discussion_r3090176510 which I am still noodling on.

davidben:
Oh forgot that hadn't been merged. Yeah, we might need more text there. Also need to give mutiple CA cosigners a think.

davidben:
Okay, so we have #218 which tries to explicitly say OCSP/CRL is OK, but that indeed reveals fun messes (CC @ilaril ), all relating to the fact that we allow multiple CA cosigners per CA:

* The CRL-DP in the certificate is part of the log entry and thus is shared between the cosigners. So either you include one URL per CA cosigner or... something
* If you did include one URL per cosigner, we need some way to tell which CA cosigner is which
* On paper, you'd think we could use a different SKID per CA cosigner, since CRLs have AKIDs. However, certificates also have AKIDs to facilitate path-building (not useful here), so we really need to make the SKID common to the CA log, not the CA cosigner

As alluded to in https://github.com/ietf-plants-wg/merkle-tree-certs/pull/218#discussion_r3157969332, it's not totally clear if we really _need_ to have multiple CA cosigners per CA. Some of us have been muttering about maybe removing this, which would simplify a bunch of things. (I've not done this in #218 yet since that's a bigger change.)

[Edit: Whoops, fixed the CC. Sorry for the spam, other Ilari!]

ilaril:
I think that CC was misdirected...

>     * The CRL-DP in the certificate is part of the log entry and thus is shared between the cosigners. So either you include one URL per CA cosigner or... something

AFAICT, that is not allowed. RFC 5280 seems to allow only one CRL per certificate ("each name describes a different mechanism to obtain the same CRL").
 
>     * If you did include one URL per cosigner, we need some way to tell which CA cosigner is which

If there was a way to attach multiple CRLs, AKID/SKID matching would seem like a natural way to do that.

[...]

PR #218, opened 2026-04-28
Define an MTC CA certificate format

### LINKED ISSUE #96: Define a MTC trust anchor serialization

davidben:
"MTC CA certificate", of course, stands for Merkle Tree Certificate certification authority certificate, because we are all certifiably good at naming things.

Closes #96.

### DIFF
--- a/draft-ietf-plants-merkle-tree-certs.md
+++ b/draft-ietf-plants-merkle-tree-certs.md
@@ -1312,6 +1312,8 @@ This section defines the X.509 Certificate {{!RFC5280}} representation of a Merk
 
 * The `extensions` field MUST contain a critical extension of type id-pe-mtcCertificationAuthority, defined below.
 
+* The subject key identifier extension ({{Section 4.2.1.2 of !RFC5280}}), if present, SHOULD be set to CA's log ID {{log-ids}}. The log ID is encoded in its binary representation, as defined in {{Section 3 of !I-D.ietf-tls-trust-anchor-ids}}.
+
 Other fields and extensions in {{!RFC5280}} apply unmodified. In particular:
 
 * The key usage extension ({{Section 4.2.1.3 of !RFC5280}}) MUST be present and assert at least the `keyCertSign` bit.
--- a/draft-ietf-plants-merkle-tree-certs.md
+++ b/draft-ietf-plants-merkle-tree-certs.md
@@ -1344,7 +1344,7 @@ MTCCertificationAuthority ::= SEQUENCE {
 
 The fields of a MTCCertificationAuthority structure are defined as follows:
 
-* `logHash` describes the hash function used in the log. For example, if the hash is SHA-256, it would be `mda-sha256` as defined in {{Section 8 of !RFC5912}}.
+* `logHash` describes the hash algorithm used in the log. For example, if the hash is SHA-256, it would be `mda-sha256` as defined in {{Section 8 of !RFC5912}}.
 
 * `cosignerID` is the CA cosigner's cosigner ID ({{cosigners}}) as a RELATIVE-OID.
 
@@ -1367,7 +1367,7 @@ This section discusses how relying parties verify Merkle Tree Certificates.
 In order to accept certificates from a Merkle Tree CA, a relying party MUST be configured with:
 
 * The log ID ({{log-ids}})
-* The log hash, e.g. SHA-256
+* The log hash algorithm, e.g. SHA-256
 * A set of supported cosigners, as pairs of cosigner ID and public key
 * A policy on which combinations of cosigners to accept in a certificate ({{trusted-cosigners}})
 * An optional list of trusted subtrees, with their hashes, that are known to be consistent with the relying party's cosigner requirements ({{trusted-subtrees}})
@@ -1377,7 +1377,7 @@ This information may be obtained, in part, from a CA certificate structure, defi
 
 * The log ID is determined from the certificate's subject.
 
-* The log hash is determined from the id-pe-mtcCertificationAuthority extension.
+* The log hash algorithm is determined from the id-pe-mtcCertificationAuthority extension.
 
 * The CA cosigner is determined from the certificate's subject public key and id-pe-mtcCertificationAuthority extension. The relying party incorporates this cosigner into its cosigner policy. {{trusted-cosigners}} gives some guidance on this.
 
--- a/draft-ietf-plants-merkle-tree-certs.md
+++ b/draft-ietf-plants-merkle-tree-certs.md
@@ -1314,7 +1314,7 @@ This section defines the X.509 Certificate {{!RFC5280}} representation of a Merk
 
 Other fields and extensions in {{!RFC5280}} apply unmodified. In particular:
 
-* The key usage extension ({{Section 4.2.1.3 of !RFC5280}}) MUST be present and assert the `keyCertSign` bit.
+* The key usage extension ({{Section 4.2.1.3 of !RFC5280}}) MUST be present and assert at least the `keyCertSign` bit.
 
 * The basic constraints extension ({{Section 4.2.1.9 of !RFC5280}}) MUST be present and set the `cA` field to TRUE.
 
--- a/draft-ietf-plants-merkle-tree-certs.md
+++ b/draft-ietf-plants-merkle-tree-certs.md
@@ -208,7 +208,7 @@ This increased overhead additionally impacts CT logs themselves. Most of a log's
 
 Additionally, as PKIs transition to shorter-lived certificates {{CABF-153}} {{CABF-SC081}}, the number of entries in the log will grow.
 
-This document introduces Merkle Tree certificates (MTCs), a new form of X.509 certificate that integrates logging with certificate issuance. Each CA maintains a log of everything it issues, signing views of the log to assert it has issued the contents. The CA signature is combined with cosignatures from other parties who verify correct operation and optionally mirror the log. These signatures, together with an inclusion proof for an individual entry, constitute a certificate.
+This document introduces Merkle Tree Certificates (MTCs), a new form of X.509 certificate that integrates logging with certificate issuance. Each CA maintains a log of everything it issues, signing views of the log to assert it has issued the contents. The CA signature is combined with cosignatures from other parties who verify correct operation and optionally mirror the log. These signatures, together with an inclusion proof for an individual entry, constitute a certificate.
 
 This achieves the following:
 
@@ -1304,17 +1304,17 @@ Proof sizes grow logarithmically, so 32 hashes, or 1024 bytes, is sufficient for
 
 ## Representing Certification Authorities
 
-This section defines X.509 Certificate {{!RFC5280}} representation for information about a Merkle Tree Certificate CA. It identifies the issuance log ({{issuance-logs}}) and one associated CA cosigner ({{certification-authority-cosigners}}). This information is encoded as follows:
+This section defines the X.509 Certificate {{!RFC5280}} representation of a Merkle Tree Certificate CA. It identifies the issuance log ({{issuance-logs}}) and one associated CA cosigner ({{certification-authority-cosigners}}). This information is encoded as follows:
 
 * The `subject` field MUST be the CA issuance log's log ID as a PKIX distinguished name, as described in {{log-ids}}.
 
-* The `subjectPublicKeyInfo` field MUST be the public key of the CA cosigner {{certification-authority-cosigners}}
+* The `subjectPublicKeyInfo` field MUST be the public key of the CA cosigner {{certification-authority-cosigners}}.
 
 * The `extensions` field MUST contain a critical extension of type id-pe-mtcCertificationAuthority, defined below.
 
 Other fields and extensions in {{!RFC5280}} apply unmodified. In particular:
 
-* The key usage extension ({{Section 4.2.1.3 of !RFC5280}} MUST be present and assert the `keyCertSign` bit.
+* The key usage extension ({{Section 4.2.1.3 of !RFC5280}}) MUST be present and assert the `keyCertSign` bit.
 
 * The basic constraints extension ({{Section 4.2.1.9 of !RFC5280}}) MUST be present and set the `cA` field to TRUE.
 
@@ -1326,7 +1326,7 @@ id-pe-mtcCertificationAuthority OBJECT IDENTIFIER ::= {
     mechanisms(5) pkix(7) pe(1) TBD }
 
 ext-mtcCertificationAuthority EXTENSION ::= {
-    SYNTAX MTCCertificateAuthority
+    SYNTAX MTCCertificationAuthority
     IDENTIFIED BY id-pe-mtcCertificationAuthority
     CRITICALITY TRUE
 }
@@ -1348,13 +1348,13 @@ The fields of a MTCCertificationAuthority structure are defined as follows:
 
 * `cosignerID` is the CA cosigner's cosigner ID ({{cosigners}}) as a RELATIVE-OID.
 
-* `cosignerSigAlg` the CA cosigner's signature algorithm ({{signature-algorithms}}).
+* `cosignerSigAlg` is the CA cosigner's signature algorithm ({{signature-algorithms}}).
 
 * `minSerial` is an integer describing the minimum allowed serial number in the issuance log ({{log-pruning}}).
 
-If this extension is present, the key described in `subjectPublicKeyInfo` MUST NOT be used to directly sign TBSCertificate structures, as in a traditional X.509 CA. Instead, it is used to sign subtrees as described in {{signature-format}}. The key MAY be used to directly sign certificate revocations lists (CRLs) {{!RFC5280}} and OCSP responses {{!RFC6960}}. CRLs and OCSP apply to Merkle Tree Certificates unchanged, though later documents MAY introduce new integrations.
+If this extension is present, the key described in `subjectPublicKeyInfo` MUST NOT be used to directly sign TBSCertificate structures, as in a traditional X.509 CA. Instead, it is used to sign subtrees as described in {{signature-format}}. The key MAY be used to directly sign certificate revocation lists (CRLs) {{!RFC5280}} and Online Certificate Status Protocol (OCSP) responses {{!RFC6960}}. CRLs and OCSP apply to Merkle Tree Certificates unchanged, though later documents MAY introduce new integrations.
 
-This extension indicates the subtree signature format defined in {{signature-format}}. If a later version of the protocol define a new format, this SHOULD be represented in CA certificates with a new extension type.
+This extension indicates the subtree signature format defined in {{signature-format}}. If a later version of the protocol defines a new format, this SHOULD be represented in CA certificates with a new extension type.
 
 A CA certificate using this format SHOULD NOT be self-signed by the Merkle Tree Certificate CA. Doing so would require writing the information in the issuance log. Instead, if used to represent a trust anchor, the certificate should be an unsigned certificate {{!RFC9925}}.
 
@@ -1366,8 +1366,8 @@ This section discusses how relying parties verify Merkle Tree Certificates.
 
 In order to accept certificates from a Merkle Tree CA, a relying party MUST be configured with:
 
-* The log hash, e.g. SHA-256
 * The log ID ({{log-ids}})
+* The log hash, e.g. SHA-256
 * A set of supported cosigners, as pairs of cosigner ID and public key
 * A policy on which combinations of cosigners to accept in a certificate ({{trusted-cosigners}})
 * An optional list of trusted subtrees, with their hashes, that are known to be consistent with the relying party's cosigner requirements ({{trusted-subtrees}})
@@ -1375,15 +1375,15 @@ In order to accept certificates from a Merkle Tree CA, a relying party MUST be c
 
 This information may be obtained, in part, from a CA certificate structure, defined in {{representing-certification-authorities}}:
 
-* The log hash is determined from the id-pe-mtcCertificationAuthority extension.
-
 * The log ID is determined from the certificate's subject.
+
+* The log hash is determined from the id-pe-mtcCertificationAuthority extension.
 
 * The CA cosigner is determined from the certificate's subject public key and id-pe-mtcCertificationAuthority extension. The relying party incorporates this cosigner into its cosigner policy. {{trusted-cosigners}} gives some guidance on this.
 
 * No trusted subtrees are directly represented by the CA certificate structure, but the relying party MAY incorporate additional out-of-band information.
 
-* The revoked indices contains the half-open range `[0, minSerial)`, but the relying party MAY incorporate additional out-of-band information.
+* The revoked indices contains the half-open range `[0, minSerial)`, but the relying party MAY incorporate additional ranges obtained via out-of-band information.
 
 ## Verifying Certificate Signatures
 
@@ -1765,13 +1765,13 @@ IANA is requested to add the following entry to the "SMI Security for PKIX Algor
 |---------|-----------------|------------|
 | TBD     | id-alg-mtcProof | [this-RFC] |
 
-# Certificate Extension
+## Certificate Extension
 
 IANA is requested to add the following entry to the "SMI Security for PKIX Certificate Extension" registry {{?RFC7299}}:
 
 | Decimal | Description                      | References |
 |---------|----------------------------------|------------|
-| TBD     | id-alg-mtcCertificationAuthority | [this-RFC] |
+| TBD     | id-pe-mtcCertificationAuthority | [this-RFC] |
 
 ## Relative Distinguished Name Attribute
 
@@ -1858,7 +1858,7 @@ id-pe-mtcCertificationAuthority OBJECT IDENTIFIER ::= {
     mechanisms(5) pkix(7) pe(1) TBD }
 
 ext-mtcCertificationAuthority EXTENSION ::= {
-    SYNTAX MTCCertificateAuthority
+    SYNTAX MTCCertificationAuthority
     IDENTIFIED BY id-pe-mtcCertificationAuthority
     CRITICALITY TRUE
 }
--- a/draft-ietf-plants-merkle-tree-certs.md
+++ b/draft-ietf-plants-merkle-tree-certs.md
@@ -1373,7 +1373,7 @@ In order to accept certificates from a Merkle Tree CA, a relying party MUST be c
 * An optional list of trusted subtrees, with their hashes, that are known to be consistent with the relying party's cosigner requirements ({{trusted-subtrees}})
 * A list of revoked ranges of indices ({{revocation-by-index}})
 
-This information may be obtained, in part, from a CA certificate structure, defined in {{representing-certification-authorities}}:
+This information may be obtained from a CA certificate structure, defined in {{representing-certification-authorities}}:
 
 * The log ID is determined from the certificate's subject.
 
@@ -1381,9 +1381,9 @@ This information may be obtained, in part, from a CA certificate structure, defi
 
 * The CA cosigner is determined from the certificate's subject public key and id-pe-mtcCertificationAuthority extension. The relying party incorporates this cosigner into its cosigner policy. {{trusted-cosigners}} gives some guidance on this.
 
-* No trusted subtrees are directly represented by the CA certificate structure, but the relying party MAY incorporate additional out-of-band information.
+* No trusted subtrees are directly represented by the CA certificate structure, but the relying party MAY incorporate trusted subtrees from out-of-band information.
 
-* The revoked indices contains the half-open range `[0, minSerial)`, but the relying party MAY incorporate additional ranges obtained via out-of-band information.
+* The revoked index ranges include the half-open range `[0, minSerial)`, but the relying party MAY incorporate additional ranges from out-of-band information.
 
 ## Verifying Certificate Signatures
 
--- a/draft-ietf-plants-merkle-tree-certs.md
+++ b/draft-ietf-plants-merkle-tree-certs.md
@@ -1350,9 +1350,9 @@ The fields of a MTCCertificationAuthority structure are defined as follows:
 
 * `cosignerSigAlg` is the CA cosigner's signature algorithm ({{signature-algorithms}}).
 
-* `minSerial` is an integer describing the minimum allowed serial number in the issuance log ({{log-pruning}}).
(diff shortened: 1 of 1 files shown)


davidben:
~~(Will send mail about this tomorrow.)~~ Or now...

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1309,21 +1302,88 @@ If a new landmark is allocated every hour, landmark-relative certificate subtree
 
 Proof sizes grow logarithmically, so 32 hashes, or 1024 bytes, is sufficient for subtrees of up to 2<sup>32</sup> (4,294,967,296) certificates.
 
+## Representing Certification Authorities
+
+This section defines X.509 Certificate {{!RFC5280}} representation for information about a Merkle Tree Certificate CA. It identifies the issuance log ({{issuance-logs}}) and one associated CA cosigner ({{certification-authority-cosigners}}). This information is encoded as follows:
+
+* The `subject` field MUST be the CA issuance log's log ID as a PKIX distinguished name, as described in {{log-ids}}.
+
+* The `subjectPublicKeyInfo` field MUST be the public key of the CA cosigner {{certification-authority-cosigners}}
```
```suggestion
* The `subjectPublicKeyInfo` field MUST be the public key of the CA cosigner {{certification-authority-cosigners}}.
```

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -208,7 +208,7 @@ This increased overhead additionally impacts CT logs themselves. Most of a log's
 
 Additionally, as PKIs transition to shorter-lived certificates {{CABF-153}} {{CABF-SC081}}, the number of entries in the log will grow.
 
-This document introduces Merkle Tree certificates, a new form of X.509 certificate that integrates logging with certificate issuance. Each CA maintains a log of everything it issues, signing views of the log to assert it has issued the contents. The CA signature is combined with cosignatures from other parties who verify correct operation and optionally mirror the log. These signatures, together with an inclusion proof for an individual entry, constitute a certificate.
+This document introduces Merkle Tree certificates (MTCs), a new form of X.509 certificate that integrates logging with certificate issuance. Each CA maintains a log of everything it issues, signing views of the log to assert it has issued the contents. The CA signature is combined with cosignatures from other parties who verify correct operation and optionally mirror the log. These signatures, together with an inclusion proof for an individual entry, constitute a certificate.
```
lukevalenta: Non-blocking nit: Not sure if we're doing upper or lowercase 'c', but we should be consistent. If we're doing lowercase, should 'tree' also be lowercase? Right now the doc has several instances of each. I guess the rule is uppercase 'Certificate' when talking about the protocol, and lowercase 'certificate' when talking about the cert objects. Maybe we just try to remove the 'Merkle Tree' prefix when talking about cert objects? Or even 'MTC certificates'
```suggestion
This document introduces Merkle Tree Certificates (MTCs), a new form of X.509 certificate that integrates logging with certificate issuance. Each CA maintains a log of everything it issues, signing views of the log to assert it has issued the contents. The CA signature is combined with cosignatures from other parties who verify correct operation and optionally mirror the log. These signatures, together with an inclusion proof for an individual entry, constitute a certificate.
```

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1309,21 +1302,88 @@ If a new landmark is allocated every hour, landmark-relative certificate subtree
...
+
+The fields of a MTCCertificationAuthority structure are defined as follows:
+
+* `logHash` describes the hash function used in the log. For example, if the hash is SHA-256, it would be `mda-sha256` as defined in {{Section 8 of !RFC5912}}.
+
+* `cosignerID` is the CA cosigner's cosigner ID ({{cosigners}}) as a RELATIVE-OID.
+
+* `cosignerSigAlg` the CA cosigner's signature algorithm ({{signature-algorithms}}).
+
+* `minSerial` is an integer describing the minimum allowed serial number in the issuance log ({{log-pruning}}).
```
lukevalenta: A parenthetical or something here might be a nice reminder for readers.
```suggestion
* `minSerial` is an integer describing the minimum allowed serial number (leaf index) in the issuance log ({{log-pruning}}).
```

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1309,21 +1302,88 @@ If a new landmark is allocated every hour, landmark-relative certificate subtree
 
 Proof sizes grow logarithmically, so 32 hashes, or 1024 bytes, is sufficient for subtrees of up to 2<sup>32</sup> (4,294,967,296) certificates.
 
+## Representing Certification Authorities
+
+This section defines X.509 Certificate {{!RFC5280}} representation for information about a Merkle Tree Certificate CA. It identifies the issuance log ({{issuance-logs}}) and one associated CA cosigner ({{certification-authority-cosigners}}). This information is encoded as follows:
```
```suggestion
This section defines the X.509 Certificate {{!RFC5280}} representation of a Merkle Tree Certificate CA. It identifies the issuance log ({{issuance-logs}}) and one associated CA cosigner ({{certification-authority-cosigners}}). This information is encoded as follows:
```

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1309,21 +1302,88 @@ If a new landmark is allocated every hour, landmark-relative certificate subtree
...
+
+* `logHash` describes the hash function used in the log. For example, if the hash is SHA-256, it would be `mda-sha256` as defined in {{Section 8 of !RFC5912}}.
+
+* `cosignerID` is the CA cosigner's cosigner ID ({{cosigners}}) as a RELATIVE-OID.
+
+* `cosignerSigAlg` the CA cosigner's signature algorithm ({{signature-algorithms}}).
+
+* `minSerial` is an integer describing the minimum allowed serial number in the issuance log ({{log-pruning}}).
+
+If this extension is present, the key described in `subjectPublicKeyInfo` MUST NOT be used to directly sign TBSCertificate structures, as in a traditional X.509 CA. Instead, it is used to sign subtrees as described in {{signature-format}}. The key MAY be used to directly sign certificate revocations lists (CRLs) {{!RFC5280}} and OCSP responses {{!RFC6960}}. CRLs and OCSP apply to Merkle Tree Certificates unchanged, though later documents MAY introduce new integrations.
```
```suggestion
If this extension is present, the key described in `subjectPublicKeyInfo` MUST NOT be used to directly sign TBSCertificate structures, as in a traditional X.509 CA. Instead, it is used to sign subtrees as described in {{signature-format}}. The key MAY be used to directly sign certificate revocation lists (CRLs) {{!RFC5280}} and Online Certificate Status Protocol (OCSP) responses {{!RFC6960}}. CRLs and OCSP apply to Merkle Tree Certificates unchanged, though later documents MAY introduce new integrations.
```

```suggestion
If this extension is present, the key described in `subjectPublicKeyInfo` MUST NOT be used to directly sign TBSCertificate structures, as in a traditional X.509 CA. Instead, it is used to sign subtrees as described in {{signature-format}}. The key MAY be used to directly sign certificate revocation lists (CRLs) {{!RFC5280}} and OCSP responses {{!RFC6960}}. CRLs and OCSP apply to Merkle Tree Certificates unchanged, though later documents MAY introduce new integrations.
```

Also, in order to sign OCSP responses, it would need to assert the digitalSignature KU. I think it's totally fine if it does that, but the combination of line 1317 and this line (saying it MAY be used) might be interpreted as implying that an MTC CA can sign OCSP even without the digitalSignature KU.

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1309,21 +1302,88 @@ If a new landmark is allocated every hour, landmark-relative certificate subtree
...
+
+This extension indicates the subtree signature format defined in {{signature-format}}. If a later version of the protocol define a new format, this SHOULD be represented in CA certificates with a new extension type.
```
```suggestion
This extension indicates the subtree signature format defined in {{signature-format}}. If a later version of the protocol defines a new format, this SHOULD be represented in CA certificates with a new extension type.
```

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1309,21 +1302,88 @@ If a new landmark is allocated every hour, landmark-relative certificate subtree
...
 # Relying Parties
 
 This section discusses how relying parties verify Merkle Tree Certificates.
 
-## Trust Anchors
+## Relying Party Configuration
 
 In order to accept certificates from a Merkle Tree CA, a relying party MUST be configured with:
 
+* The log hash, e.g. SHA-256
```
lukevalenta: Just to clarify that this is the algorithm, not a root hash or something.
```suggestion
* The log hash algorithm, e.g. SHA-256
```

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1309,21 +1302,88 @@ If a new landmark is allocated every hour, landmark-relative certificate subtree
...
 * The log ID ({{log-ids}})
```
aarongable: nit:
```suggestion
* The log ID ({{log-ids}})
* The log hash, e.g. SHA-256
```

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1309,21 +1302,88 @@ If a new landmark is allocated every hour, landmark-relative certificate subtree
...
 * A policy on which combinations of cosigners to accept in a certificate ({{trusted-cosigners}})
 * An optional list of trusted subtrees, with their hashes, that are known to be consistent with the relying party's cosigner requirements ({{trusted-subtrees}})
 * A list of revoked ranges of indices ({{revocation-by-index}})
 
-[[TODO: Define some representation for this. In a trust anchor, there's a lot of room for flexibility in what the client stores. In principle, we could even encode some of this information in an X.509 intermediate certificate, if an application wishes to use this with a delegation model with intermediates, though the security story becomes more complex. Decide how/whether to do that.]]
+This information may be obtained, in part, from a CA certificate structure, defined in {{representing-certification-authorities}}:
+
+* The log hash is determined from the id-pe-mtcCertificationAuthority extension.
+
+* The log ID is determined from the certificate's subject.
```
aarongable: same nit:
```suggestion
* The log ID is determined from the certificate's subject.

* The log hash is determined from the id-pe-mtcCertificationAuthority extension.
```

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1378,13 +1438,15 @@ Authenticity:
 Transparency:
 : The relying party only accepts entries that are publicly accessible, so that monitors, particularly the subject of the certificate, can notice any unauthorized certificates
 
-Relying parties SHOULD ensure authenticity by requiring a signature from the most recent CA cosigner key. If the CA is transitioning from an old to new key, the relying party SHOULD accept both keys until certificates that predate the new key expire. This is analogous to the signature in a traditional X.509 certificate.
+Relying parties SHOULD ensure authenticity by requiring a signature from the most recent CA cosigner key. This is analogous to the signature in a traditional X.509 certificate. If the relying party obtains CA information from a CA certificate, the CA cosigner key is determined as in {{relying-party-configuration}}. If the CA is transitioning from an old to new key, the relying party SHOULD accept both keys until certificates that predate the new key expire.
```
aarongable, line 1447: Shouldn't we just ban key rotations? If you need to use a new key, start a new tree.

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1309,21 +1302,88 @@ If a new landmark is allocated every hour, landmark-relative certificate subtree
...
+
+* The log hash is determined from the id-pe-mtcCertificationAuthority extension.
+
+* The log ID is determined from the certificate's subject.
+
+* The CA cosigner is determined from the certificate's subject public key and id-pe-mtcCertificationAuthority extension. The relying party incorporates this cosigner into its cosigner policy. {{trusted-cosigners}} gives some guidance on this.
+
+* No trusted subtrees are directly represented by the CA certificate structure, but the relying party MAY incorporate additional out-of-band information.
+
+* The revoked indices contains the half-open range `[0, minSerial)`, but the relying party MAY incorporate additional out-of-band information.
```
```suggestion
* The revoked indices contains the half-open range `[0, minSerial)`, but the relying party MAY incorporate additional ranges obtained via out-of-band information.
```

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1782,6 +1853,23 @@ at-trustAnchorID ATTRIBUTE ::= {
     IDENTIFIED BY id-rdna-trustAnchorID
 }
 
+id-pe-mtcCertificationAuthority OBJECT IDENTIFIER ::= {
+    iso(1) identified-organization(3) dod(6) internet(1) security(5)
+    mechanisms(5) pkix(7) pe(1) TBD }
+
+ext-mtcCertificationAuthority EXTENSION ::= {
+    SYNTAX MTCCertificateAuthority
```
```suggestion
    SYNTAX MTCCertificationAuthority
```

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1309,21 +1302,88 @@ If a new landmark is allocated every hour, landmark-relative certificate subtree
...
+
+The id-pe-mtcCertificationAuthority is defined below. This extension indicates that the subject of the certificate is a CA that issues Merkle Tree Certificates. If present, it MUST be marked as critical.
+
+~~~asn.1
+id-pe-mtcCertificationAuthority OBJECT IDENTIFIER ::= {
+    iso(1) identified-organization(3) dod(6) internet(1) security(5)
+    mechanisms(5) pkix(7) pe(1) TBD }
+
+ext-mtcCertificationAuthority EXTENSION ::= {
+    SYNTAX MTCCertificateAuthority
```
```suggestion
    SYNTAX MTCCertificationAuthority
```

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1703,6 +1765,14 @@ IANA is requested to add the following entry to the "SMI Security for PKIX Algor
 |---------|-----------------|------------|
 | TBD     | id-alg-mtcProof | [this-RFC] |
 
+# Certificate Extension
```
```suggestion
## Certificate Extension
```

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1703,6 +1765,14 @@ IANA is requested to add the following entry to the "SMI Security for PKIX Algor
...
+
+IANA is requested to add the following entry to the "SMI Security for PKIX Certificate Extension" registry {{?RFC7299}}:
+
+| Decimal | Description                      | References |
+|---------|----------------------------------|------------|
+| TBD     | id-alg-mtcCertificationAuthority | [this-RFC] |
```
```suggestion
| TBD     | id-pe-mtcCertificationAuthority | [this-RFC] |
```

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1309,21 +1302,88 @@ If a new landmark is allocated every hour, landmark-relative certificate subtree
...
+
+* The `subject` field MUST be the CA issuance log's log ID as a PKIX distinguished name, as described in {{log-ids}}.
+
+* The `subjectPublicKeyInfo` field MUST be the public key of the CA cosigner {{certification-authority-cosigners}}
+
+* The `extensions` field MUST contain a critical extension of type id-pe-mtcCertificationAuthority, defined below.
+
+Other fields and extensions in {{!RFC5280}} apply unmodified. In particular:
+
+* The key usage extension ({{Section 4.2.1.3 of !RFC5280}} MUST be present and assert the `keyCertSign` bit.
```
```suggestion
* The key usage extension ({{Section 4.2.1.3 of !RFC5280}}) MUST be present and assert the `keyCertSign` bit.
```

draft-ietf-plants-merkle-tree-certs.md
lukevalenta: Stray backtick after 'zero' at the end.
```suggestion
1. Otherwise, check that the MTCProof's `signatures` contain a sufficient set of valid signatures from cosigners to satisfy the relying party's cosigner requirements ({{trusted-cosigners}}). Unrecognized cosigners MUST be ignored. Signatures are verified as described in {{signature-format}}. Reconstruct the CosignedMessage from MTCProof's `start` and `end`, the cosigner ID for `cosigner_name`, the log ID for `log_origin`, `expected_subtree_hash` for `subtree_hash`, and `timestamp` set to zero.
```

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1309,21 +1302,88 @@ If a new landmark is allocated every hour, landmark-relative certificate subtree
...
+}
+~~~
+
+The fields of a MTCCertificationAuthority structure are defined as follows:
+
+* `logHash` describes the hash function used in the log. For example, if the hash is SHA-256, it would be `mda-sha256` as defined in {{Section 8 of !RFC5912}}.
+
+* `cosignerID` is the CA cosigner's cosigner ID ({{cosigners}}) as a RELATIVE-OID.
+
+* `cosignerSigAlg` the CA cosigner's signature algorithm ({{signature-algorithms}}).
```
```suggestion
* `cosignerSigAlg` is the CA cosigner's signature algorithm ({{signature-algorithms}}).
```

draft-ietf-plants-merkle-tree-certs.md
```suggestion
This extension indicates the subtree signature format defined in {{signature-format}}. If a later version of the protocol defines a new format, this SHOULD be represented in CA certificates with a new extension type.
```

lukevalenta (approved):
Some small cleanup/nits (feel free to ignore any you feel don't apply), but otherwise LGTM!

draft-ietf-plants-merkle-tree-certs.md
davidben: Done. Used "entry index" just to avoid fussing over whether it's always a "leaf".

draft-ietf-plants-merkle-tree-certs.md
davidben: Applied manually since it conflicted with other comments.

draft-ietf-plants-merkle-tree-certs.md
davidben: (Applied via another parallel comment.)

draft-ietf-plants-merkle-tree-certs.md
davidben: > Also, in order to sign OCSP responses, it would need to assert the digitalSignature KU. I think it's totally fine if it does that, but the combination of line 1317 and this line (saying it MAY be used) might be interpreted as implying that an MTC CA can sign OCSP even without the digitalSignature KU.

Ah good call. I added a "if permitted by other X.509 constraints such as the key usage extension" and then removed the sentence afterwards because it didn't seem to flow.

draft-ietf-plants-merkle-tree-certs.md
davidben: Quite possibly. @jdeblasio and I have been muttering about this. I left it alone here just because it seems orthogonal.

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1309,21 +1302,88 @@ If a new landmark is allocated every hour, landmark-relative certificate subtree
...
+
+* The `subject` field MUST be the CA issuance log's log ID as a PKIX distinguished name, as described in {{log-ids}}.
+
+* The `subjectPublicKeyInfo` field MUST be the public key of the CA cosigner {{certification-authority-cosigners}}.
+
+* The `extensions` field MUST contain a critical extension of type id-pe-mtcCertificationAuthority, defined below.
+
+Other fields and extensions in {{!RFC5280}} apply unmodified. In particular:
+
+* The key usage extension ({{Section 4.2.1.3 of !RFC5280}}) MUST be present and assert the `keyCertSign` bit.
```
robstradling: [Emphasis mine] RFC5280 says that *"The keyCertSign bit is asserted when the subject public key is used for verifying signatures **on** public key certificates"*, and X.509 (10/2019) says *"keyCertSign: for verifying a CA's signature **on** public-key certificates"*.
Whereas this PR says that *"the key described in `subjectPublicKeyInfo` MUST NOT be used to **directly sign** TBSCertificate structures"*.

ISTM that "on" equates to "directly sign", meaning that it's inappropriate to assert `keyCertSign` in an MTC CA certificate.  Instead, the appropriate key usage bit to use here is `digitalSignature`, which per RFC5280 *"is used for verifying digital signatures, other than signatures **on** certificates (bit 5) and CRLs (bit 6)"*.

robstradling: Also, given that *"the key MAY be used to directly sign certificate revocation lists (CRLs) {{!RFC5280}} and Online Certificate Status Protocol (OCSP) responses {{!RFC6960}}"*, it's necessary to permit `digitalSignature` (for direct signing of OCSP responses) and `cRLSign` (for direct signing of CRLs).

robstradling:
```suggestion
* The key usage extension ({{Section 4.2.1.3 of !RFC5280}}) MUST be present and assert the `digitalSignature` bit, and MAY also assert the `cRLSign` bit.
```

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1309,21 +1302,88 @@ If a new landmark is allocated every hour, landmark-relative certificate subtree
...
+
+* The basic constraints extension ({{Section 4.2.1.9 of !RFC5280}}) MUST be present and set the `cA` field to TRUE.
```
robstradling, line 1321: [Emphasis mine] RFC5280 says that *"The cA boolean indicates whether the certified public key may be used to verify **certificate signatures**"*, and X.509 (10/2019) says that *"The cA component with the value TRUE indicates that the certified public key may be used to verify public-key **certificate signatures**"*.

However, "certificate signatures" (signatures calculated over individual TBSCertificate structures) are not a thing in an MTC world, so ISTM that it's inappropriate to set the `cA` field to TRUE in an MTC CA certificate.

robstradling:
```suggestion
* The basic constraints extension ({{Section 4.2.1.9 of !RFC5280}}) MUST be present, MUST NOT set the `cA` field to TRUE, and MUST NOT include a `pathLenConstraint`.
```

aarongable: I disagree that the phrase "certificate signatures" exclusively applies to signatures computed over the bytes of (and only the bytes of) a TBSCertificate. A certificate signature is whatever is in the signature field of a certificate. MTCerts will contain signatures, those signatures will be computed over (a transform of) their TBSCertificate bytes (and some other bytes), and the MTCA's pubkey will be used to verify those signatures. Thus it qualifies as a CA per the language of RFC 5280 and X.509.

draft-ietf-plants-merkle-tree-certs.md
aarongable: The signature "on" a public key certificate is the signature contained in its signature field. If this said that "the keyCertSign bit is asserted with the private key corresponding to the subject public key is used for producing signatures over TBSCertificates", then I'd agree with you, but that's not what the standards say -- they are phrased in terms of _verifying_ signatures, and an MTCA's public key _is_ used to verify the signature contained in a a Merkle Tree Cert.

I feel less strongly about this than about the one below, however, because I agree that "digitalSignature" is _also_ a thing that these keypairs are clearly doing.

davidben: It definitely should be `keyCertSign`. Practically, if we change it, now everyone's X.509 logic needs to be changed to say that sometimes the signature on the certificate uses `keyCertSign` and sometimes it uses `digitalSignature`.

But also semantically, `keyCertSign` is correct. New X.509 signature algorithms still use the `keyCertSign` bit and this is "just" a new signature algorithm that signs TBSCertificates in a different way.

Whether they also assert other bits is a function of what else those keys are doing. The intent was to not constrain that in this sentence. I.e. "assert the keyCertSign bit" does not preclude other bits. I think that's consistent with RFC 5280? None of this text precludes other bits from _also_ being asserted:

* https://www.rfc-editor.org/rfc/rfc5280#:~:text=extension%2C%20if%20present%2C-,asserts%20the%0A%20%20%20keyCertSign%20bit,-(Section%204.2.1.3
* https://www.rfc-editor.org/rfc/rfc5280#:~:text=For%0A%20%20%20example%2C%20when%20an,bit%20would%20be%20asserted.
* https://www.rfc-editor.org/rfc/rfc5280#:~:text=The%20keyCertSign%20bit%20is%20asserted%20when%20the%20subject%20public%20key%20is%0A%20%20%20%20%20%20used%20for%20verifying%20signatures%20on%20public%20key%20certificates.%20%20If%20the%0A%20%20%20%20%20%20keyCertSign%20bit%20is%20asserted%2C%20then%20the%20cA%20bit%20in%20the%20basic%0A%20%20%20%20%20%20constraints%20extension%20(Section%204.2.1.9)%20MUST%20also%20be%20asserted.
* And lots of others

Really this whole bit of text was not meant to introduce new constraints, just clarify what it means when we say that the rest of X.509 applies as-is. :-)

draft-ietf-plants-merkle-tree-certs.md
davidben: (Ditto from other thread.)

draft-ietf-plants-merkle-tree-certs.md
davidben: (Added an 'at least' per some offline discussion, to make this reading more obvious.)

ilaril:
Since there can be multiple keys for a single name, I think CRL signing would require the CA certificates to have Subject Key Identifier. While using issuer/serial in CRL would be in theory possible, I suspect that is very poorly supported in practice.

draft-ietf-plants-merkle-tree-certs.md
robstradling: Thanks @davidben.  Per that offline discussion: 'at least' works for me, and you and Aaron persuaded me that `keyCertSign` is appropriate here.  (A signature "on" a certificate just means that the signature is attached to the certificate; it doesn't dictate the nature of the signature algorithm.  And `keyCertSign` really should've been called `keyCertVerify`, but it's a bit late for that now!)
This sub-thread can be marked as resolved as far as I'm concerned.

draft-ietf-plants-merkle-tree-certs.md
robstradling: You've convinced me.  The CA cert's SPKI, and the phrases "certificate signature" and "signature on certificate", don't dictate the nature of the signature algorithm on certs issued by that CA.  So if we want to say that MTC uses a funny kind of hash/sig alg, then we absolutely can.
Thanks for helping me through this thought process.  This sub-thread can be marked as resolved as far as I'm concerned.

davidben:
> Since there can be multiple keys for a single name, I think CRL signing would require the CA certificates to have Subject Key Identifier. While using issuer/serial in CRL would be in theory possible, I suspect that is very poorly supported in practice.

Ah yup, that is indeed a mess. I've added a TODO into this PR referencing #214 and left some notes there.

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1309,21 +1302,94 @@ If a new landmark is allocated every hour, landmark-relative certificate subtree
...
+
+* `logHash` describes the hash algorithm used in the log. For example, if the hash is SHA-256, it would be `mda-sha256` as defined in {{Section 8 of !RFC5912}}.
+
+* `cosignerID` is the CA cosigner's cosigner ID ({{cosigners}}) as a RELATIVE-OID.
+
+* `cosignerSigAlg` is the CA cosigner's signature algorithm ({{signature-algorithms}}).
+
+* `minSerial` is an integer describing the minimum allowed serial number (entry index) in the issuance log ({{log-pruning}}).
+
+If this extension is present, the key described in `subjectPublicKeyInfo` MUST NOT be used to directly sign TBSCertificate structures, as in a traditional X.509 CA. Instead, it is used to sign subtrees as described in {{signature-format}}. However, the key MAY be used to directly sign certificate revocation lists (CRLs) {{!RFC5280}} and Online Certificate Status Protocol (OCSP) responses {{!RFC6960}}, if permitted by other X.509 constraints such as the key usage extension.
```
robstradling, line 1359: [RFC6960 section 2.2](https://datatracker.ietf.org/doc/html/rfc6960#section-2.2) states that *"There is one basic type of OCSP response that MUST be supported by all OCSP servers and clients"* and that that mandatory-to-implement response type may be signed indirectly by a CA Designated Responder certificate (a leaf cert with the id-kp-OCSPSigning EKU).
However, when this PR says *"the key described in `subjectPublicKeyInfo` MUST NOT be used to directly sign TBSCertificate structures"* it forbids issuance of CA Designated Responder certificates.
Forbidding something that's mandatory-to-implement seems problematic.

Might it be possible to allow an exception (to the MUST NOT rule) such that "the key described in `subjectPublicKeyInfo`" MAY directly sign CA Designated Responder certificates?

bwesterb (approved):

bwesterb:
Maybe we should say something in the security considerations about the domain separation of what the private key signs? Maybe... we should actually use the context parameter for the non standard subtree use case?

draft-ietf-plants-merkle-tree-certs.md
aarongable: I don't read RFC 6960 as saying that CA Designated Responders are mandatory-to-implement. _If_ you are implementing RFC 6960 (which not all CAs do anymore) then you have to have a way to produce OCSP responses. But that can be either via Designated Responders, or via signing the OCSP response directly. This language merely makes it so that using a Designated Responder is forbidden.

davidben: I'm confused, why would we need an exception here? A CA Designated Responder is just a certificate your CA issues, right? This whole document is about how these keys issue certificates. You would just issue certificates the way this document says to, no? (I.e. by putting them in the log, etc.)

robstradling: @davidben OK, I can't argue with that (in which case it doesn't actually matter if @aarongable's reading is correct or not).
I suppose some OCSP response verifier code might even get MTC support for free (if the underlying cert chain handling code is shared with other use cases).

davidben:
> Maybe we should say something in the security considerations about the domain separation of what the private key signs?

Ah yeah @sgmenda was working on that over in #210. Let's maybe do that as a follow-up. I'll file a bug so we don't forget.

> Maybe... we should actually use the context parameter for the non standard subtree use case?

That would make it incompatible with tlog-cosignature. We could change tlog-cosignature, but I think a special prefix is fine. That makes it easier to specify in light of X.509, which defines the ML-DSA sigalg AlgIDs to not use context strings, and avoids mishaps where someone applies this to an algorithm that doesn't use context strings and loses domain separation.

(I know of one popular cryptographic library, whose signing API silently drops the context parameter for any ctx-less algorithm.)

Issue #152, opened 2025-10-15
Be more opinionated about ID allocation


davidben:
Right now we have the following IDs, to denote the various different animals:

1. The log has an ID
2. Each cosigner has an ID, including the cosigner operated by the log (CA) operator
3. The landmark sequence has an ID arc

For now, they refer to slightly different things, and we allow for some generality that we may not need:

* A CA can have multiple cosigners for a single log
* A landmark sequence could be allocated by an entity other than the CA, if the CA finds this difficult

It is not clear how many of these we need. The first is somewhat related to the question of how best to manage key rotation and log rotation. The second is... almost certainly not needed. I think it just came out of building atop tlog in the most orthogonal ways possible.

And then there's the question of whether these IDs are allowed to overlap or must they be different. I think that comes from designs of cosigner negotiation.

Once the dust settles on the rest of the details, we should take a pass and:
* Decide if any IDs should overlap
* If IDs don't overlap, perhaps ask the CA operator to pick an ID arc, and then the protocol determines the substructure for you.

davidben:
Via chatting with @nharper: whatever our ID allocation strategy is, we should have a section that summarizes all the IDs that one CA has.

PR #220, opened 2026-05-05
Only one CA cosigner per CA

### LINKED ISSUE #212: Identify standalone certificates by cosigner ID
It's not full cosigner negotiation (#199), but initially I think we can get by with:

* An analog of the CT enforcement shut-off for transparency cosigners
* Negotiating the CA cosigner as if it were a trust anchor

More specifically, suppose you have some CA whose issuance log (or group of issuance logs post #26) named L and has CA cosigners C1, C2, C3, etc., each of which sign L. A RP that specifically wants cosigner C2 would advertise a trust anchor ID C2 (or a trust anchor group containing C2). That would tell the AP that should send a certificate from L with cosigner C2 and that it can skip cosignatures C1 and C2. The RP _may_ yet require other witness/mirror cosigners (say, one of W1, W2, W3, W4, ...), so you should still send those. The negotiation scheme says nothing (yet) about that.

And then if either the RP or AP is incapable of negotiation and you decide that this is the fallback certificate, you send a jumbo certificate with every cosigner you've got, because you've got nothing better you can do.

### LINKED ISSUE #214: Make sure OCSP and CRLs can *actually* be applied as-is

davidben:
(Or at least one distinguished one with the same name as the CA / log.)

When we first wrote down the design, it was still pretty early in figuring out some of the details in MTCs, so we left many things flexible. In particular, I was inspired by how tlog-checkpoint just carried a collection of signatures, and the format wasn't terribly opinionated about what those signatures were. This gave a very nice base to express all manner of cosigner policies, etc.

But as we integrate more into X.509 proper, this becomes more challenging. An X.509 CA certificate has just the one slot for a public key. It has just the one SKID value that gets reflected into just the one downstream AKID value. There is just the one set of OCSP and CRL URLs, which serve objects with just the one slot for a signature.

While it might have been possible to engineer around this (e.g. perhaps we define a new signature algorithm for OCSP and CRLs), it seems simplest to drop this. More specifically, we say that a CA is expected to have one CA cosigner whose name matches the log ID, and that's the one used in all these X.509 integrations.

Removing this does remove what might have been a handy capability: juggling multiple CA keys without interrupting the identity of the CA. However, since sending an extra ML-DSA signature is so prohibitively expensive, we would have needed a negotiation story for cosignatures to use this anyway. Gluing all those cosignatures together would really only have been useful as a fallback when those negotiation schemes fail.

Without this, our story is instead what it would have been with traditional X.509: if you need to juggling another key, go up a layer and juggle CAs, with all the strategies we have and may need to develop for managing that.

In return, we address several open issues:

* Fixes #212 by making the issue moot. There is still value in general cosignature negotiation for the transparency-enforcing signatures, but it is less dire to solve in the first version.

* Fixes #214 by removing the conflicts around which key to use.

* CAs have one less ID to allocate, towards issue #152. That one isn't done yet. We still need to remove the landmark sequence base_id concept.

### DIFF
--- a/draft-ietf-plants-merkle-tree-certs.md
+++ b/draft-ietf-plants-merkle-tree-certs.md
@@ -883,7 +883,7 @@ Unlike {{?RFC6962}} and {{?RFC9162}}, an issuance log does not have a public sub
 
 A snapshot of the log is known as a *checkpoint*. A checkpoint is identified by its *tree size*, that is the number of elements committed to the log at the time. Its contents can be described by the Merkle Tree Hash ({{Section 2.1.1 of !RFC9162}}) of entries zero through `tree_size - 1`.
 
-Cosigners ({{cosigners}}) sign assertions about the state of the issuance log. A Merkle Tree CA operates a combination of an issuance log and CA cosigner ({{certification-authority-cosigners}}) that authenticate the log state and certifies the contents. External cosigners may also be deployed to assert correct log operation or provide other services to relying parties ({{trusted-cosigners}}).
+Cosigners ({{cosigners}}) sign assertions about the state of the issuance log. A Merkle Tree CA operates a combination of an issuance log and CA cosigner ({{certification-authority-cosigners}}) that authenticates the log state and certifies the contents. External cosigners may also be deployed to assert correct log operation or provide other services to relying parties ({{trusted-cosigners}}).
 
 ## Log Parameters
 
--- a/draft-ietf-plants-merkle-tree-certs.md
+++ b/draft-ietf-plants-merkle-tree-certs.md
@@ -883,7 +883,7 @@ Unlike {{?RFC6962}} and {{?RFC9162}}, an issuance log does not have a public sub
 
 A snapshot of the log is known as a *checkpoint*. A checkpoint is identified by its *tree size*, that is the number of elements committed to the log at the time. Its contents can be described by the Merkle Tree Hash ({{Section 2.1.1 of !RFC9162}}) of entries zero through `tree_size - 1`.
 
-Cosigners ({{cosigners}}) sign assertions about the state of the issuance log. A Merkle Tree CA operates a combination of an issuance log and one or more CA cosigners ({{certification-authority-cosigners}}) that authenticate the log state and certifies the contents. External cosigners may also be deployed to assert correct log operation or provide other services to relying parties ({{trusted-cosigners}}).
+Cosigners ({{cosigners}}) sign assertions about the state of the issuance log. A Merkle Tree CA operates a combination of an issuance log and CA cosigner ({{certification-authority-cosigners}}) that authenticate the log state and certifies the contents. External cosigners may also be deployed to assert correct log operation or provide other services to relying parties ({{trusted-cosigners}}).
 
 ## Log Parameters
 
@@ -1059,7 +1059,7 @@ Other documents or deployments MAY define other signature schemes and formats. L
 
 ## Certification Authority Cosigners
 
-A *CA cosigner* is a cosigner ({{cosigners}}) that certifies the contents of a log.
+A *CA cosigner* is a cosigner ({{cosigners}}) that certifies the contents of a log. Each CA MUST operate a CA cosigner whose cosigner ID is the same as the log ID ({{log-ids}}). A CA cosigner MUST NOT sign subtreese for logs other than its corresponding issuance log.
 
 When a CA cosigner signs a subtree, it makes the additional statement that it has certified each entry in the subtree. For example, a domain-validating CA states that it has performed domain validation for each entry, at some time consistent with the entry's validity dates. CAs are held responsible for every entry in every subtree they sign. Proving an entry is included ({{subtree-inclusion-proofs}}) in a CA-signed subtree is sufficient to prove the CA certified it.
 
@@ -1070,9 +1070,7 @@ What it means to certify an entry depends on the entry type:
 
 Entries are extensible. Future documents MAY define `type` values and what it means to certify them. A CA MUST NOT sign a subtree if it contains an entry with `type` that it does not recognize. Doing so would certify that the CA has validated the information in some not-yet-defined entry format. {{new-log-entry-types}} further discusses security implications of new formats.
 
-A CA operator MAY operate multiple CA cosigners that all certify the same log in parallel. This may be useful when, e.g., rotating CA keys. In this case, each CA instance MUST have a distinct name. The CA operator's ACME server can return all CA cosignatures together in a single certificate, with the application protocol selecting the cosignatures to use. {{use-in-tls}} describes how this is done in TLS {{!RFC8446}}.
-
-If the CA operator additionally operates a traditional X.509 CA, that CA key MUST be distinct from any Merkle Tree CA cosigner keys.
+If the CA operator additionally operates a traditional X.509 CA, that CA key MUST be distinct from the Merkle Tree CA cosigner key.
 
 ## Publishing Logs
 
@@ -1304,7 +1302,7 @@ Proof sizes grow logarithmically, so 32 hashes, or 1024 bytes, is sufficient for
 
 ## Representing Certification Authorities
 
-This section defines the X.509 Certificate {{!RFC5280}} representation of a Merkle Tree Certificate CA. It identifies the issuance log ({{issuance-logs}}) and one associated CA cosigner ({{certification-authority-cosigners}}). This information is encoded as follows:
+This section defines the X.509 Certificate {{!RFC5280}} representation of a Merkle Tree Certificate CA. It identifies the issuance log ({{issuance-logs}}) and the associated CA cosigner ({{certification-authority-cosigners}}). This information is encoded as follows:
 
 * The `subject` field MUST be the CA issuance log's log ID as a PKIX distinguished name, as described in {{log-ids}}.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1059,7 +1059,7 @@ Other documents or deployments MAY define other signature schemes and formats. L
 
 ## Certification Authority Cosigners
 
-A *CA cosigner* is a cosigner ({{cosigners}}) that certifies the contents of a log.
+A *CA cosigner* is a cosigner ({{cosigners}}) that certifies the contents of a log. Each CA MUST operate a CA cosigner whose cosigner ID is the same as the log ID ({{log-ids}}). A CA cosigner MUST NOT sign subtreese for logs other than its corresponding issuance log.
```
nharper: nit: typo in subtrees

davidben: Whoops, fixed!

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -883,7 +883,7 @@ Unlike {{?RFC6962}} and {{?RFC9162}}, an issuance log does not have a public sub
 
 A snapshot of the log is known as a *checkpoint*. A checkpoint is identified by its *tree size*, that is the number of elements committed to the log at the time. Its contents can be described by the Merkle Tree Hash ({{Section 2.1.1 of !RFC9162}}) of entries zero through `tree_size - 1`.
 
-Cosigners ({{cosigners}}) sign assertions about the state of the issuance log. A Merkle Tree CA operates a combination of an issuance log and one or more CA cosigners ({{certification-authority-cosigners}}) that authenticate the log state and certifies the contents. External cosigners may also be deployed to assert correct log operation or provide other services to relying parties ({{trusted-cosigners}}).
+Cosigners ({{cosigners}}) sign assertions about the state of the issuance log. A Merkle Tree CA operates a combination of an issuance log and CA cosigner ({{certification-authority-cosigners}}) that authenticate the log state and certifies the contents. External cosigners may also be deployed to assert correct log operation or provide other services to relying parties ({{trusted-cosigners}}).
```
```suggestion
Cosigners ({{cosigners}}) sign assertions about the state of the issuance log. A Merkle Tree CA operates a combination of an issuance log and CA cosigner ({{certification-authority-cosigners}}) that authenticates the log state and certifies the contents. External cosigners may also be deployed to assert correct log operation or provide other services to relying parties ({{trusted-cosigners}}).
```

lukevalenta (approved):

ilaril:
Might make sense to require the CA cosigner to be the first one in all stand-alone certificates.

And this does not quite fix all X.509 integration issues, as there are known extensions that hash certificates (and there might be dark extensions that do some other "creative" things).

davidben:
> Might make sense to require the CA cosigner to be the first one in all stand-alone certificates.

Hmm. Is the idea just that a simple verifier doesn't have to write a for loop? I mean, I don't mind rigidity in general, but if I were to pick a rule, I would probably say that cosigner IDs must be lexicographically sorted, so it's each to check for duplicates, etc.

> And this does not quite fix all X.509 integration issues, as there are known extensions that hash certificates (and there might be dark extensions that do some other "creative" things).

Oh? Which extensions did you have in mind?

I don't think we have to make every hypothetical dark extension interact with every other one, because that's not really possible. For example, two extensions that both believe they are computed over every other extension (think embedded SCTs) cannot interoperate. I think a more practical bar for MTC, and others, is that extensions that one might plausibly wish to use together will work. It's always an option to say "well, I guess you can't use those two features together".

ilaril:
> > Might make sense to require the CA cosigner to be the first one in all stand-alone certificates.
> 
> Hmm. Is the idea just that a simple verifier doesn't have to write a for loop? I mean, I don't mind rigidity in general, but if I were to pick a rule, I would probably say that cosigner IDs must be lexicographically sorted, so it's each to check for duplicates, etc.

Well, that could work too.

> > And this does not quite fix all X.509 integration issues, as there are known extensions that hash certificates (and there might be dark extensions that do some other "creative" things).
> 
> Oh? Which extensions did you have in mind?

Ignoring stuff where it is not clear why anyone would use those with MTCs (e.g., v1 SCTs), there is the delegated credentials stuff (TLS extension 34, gated by X.509 DelegationUsage extension).

> I don't think we have to make every hypothetical dark extension interact with every other one, because that's not really possible. For example, two extensions that both believe they are computed over every other extension (think embedded SCTs) cannot interoperate. I think a more practical bar for MTC, and others, is that extensions that one might plausibly wish to use together will work. It's always an option to say "well, I guess you can't use those two features together".

That dark extension was more about there possibly being other ways (besides hashing certificates) extensions could cause trouble. Not sure if X.509 has any extensions with "interesting" effects on EE certificates (there definitely are those kind of extensions for CA certificates).

davidben:
> Well, that could work too.

Filed https://github.com/ietf-plants-wg/merkle-tree-certs/issues/221

> there is the delegated credentials stuff (TLS extension 34, gated by X.509 DelegationUsage extension).

Ah fun. Specifically that the signature is computed over the whole certificate? Yeah, I guess, as things stand right now, you would need to go back and get a new version of the DC for each variant of the certificate that you've got. That is... not _insurmountable_ but definitely a huge pain. But yeah, a more ideal version would probably have signed over something else. :-/

(Aside: I do think the value of DCs will go down with PQC because the benefits of delegation now come at a much, much higher bandwidth price. Not to say this is now _never_ worth the trade-off. But if you increase the costs of something and keep the benefit the same, you should expect it to lose to other solutions more often.)

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1059,7 +1059,7 @@ Other documents or deployments MAY define other signature schemes and formats. L
 
 ## Certification Authority Cosigners
 
-A *CA cosigner* is a cosigner ({{cosigners}}) that certifies the contents of a log.
+A *CA cosigner* is a cosigner ({{cosigners}}) that certifies the contents of a log. Each CA MUST operate a CA cosigner whose cosigner ID is the same as the log ID ({{log-ids}}). A CA cosigner MUST NOT sign subtrees for logs other than its corresponding issuance log.
```
```suggestion
A *CA cosigner* is a cosigner ({{cosigners}}) that certifies the contents of a log. Each CA MUST operate a CA cosigner whose cosigner ID is the same as the log ID ({{log-ids}}). A CA cosigner MUST NOT sign checkpoints or subtrees for logs other than its corresponding issuance log.
```
I'm not sure if this is actually necessary, since checkpoint signatures don't show up in certs directly. But it's in line with both:
- the statement elsewhere that a Mirror must use a different key from a CA; and
- the basic idea that this CA key and only this CA key should be signing anything to do with this log.

aarongable (approved):
LGTM with nit:

draft-ietf-plants-merkle-tree-certs.md
davidben: Checkpoints are also a subset of subtrees anyway, but being explicit SGTM.

aarongable: Oh that's a good point. But even though checkpoints meet the definition of subtrees, the signed-over data is different in each case (subtree tlog notes have `subtree x y` as their second line, while checkpoint notes have just an integer size as their second line) so I think they're technically different objects.

davidben: Oh! This is just because of a series of different decisions made across different ecosystems at different times. tl;dr: They are actually the same. Pay no attention to the person^Wsigned note behind the curtain.

In MTC, the only signatures are binary. Specifically it signs over this structure: https://ietf-plants-wg.github.io/merkle-tree-certs/draft-ietf-plants-merkle-tree-certs.html#name-signature-format

tlog originally used this textual [signed note](https://c2sp.org/signed-note) format. I'm not sure exactly the history (Go sumdb?). And yeah, you're right that the originally Ed25519 [tlog-checkpoint](https://c2sp.org/tlog-checkpoint) note signature algorithm would have you sign over a text thing.

We don't really want to have MTCs sign over a text thing. Among other things, that would mean that MTC cert verifiers need a base64 encoder embedded in them. So instead we can use this trick, first done in the [Static CT API](https://c2sp.org/static-ct-api), where you say that the _actual_ signature is over whatever other representation, and the signature viewed as a signed note signature technically transforms the note first. It's all very goofy.

This is now moot because we're aligning PQC tlog and MTC on a single binary signed message, specified both [in tlog](https://github.com/C2SP/C2SP/blob/main/tlog-cosignature.md#ml-dsa-44-signed-message) and in [MTC](https://ietf-plants-wg.github.io/merkle-tree-certs/draft-ietf-plants-merkle-tree-certs.html#name-signature-format).

This signature input can represent subtrees and checkpoints. The checkpoint representation is just a special case of the subtree representation, where `start` is zero.

When those signatures are served from a tlog checkpoint, they're still dressed up in a signed note. Viewed as a signature over the note, you first convert from the text thing to that binary format, and then sign that. But the input to the private key signature function is that singular binary format. Any reasonable notion of "what the key signed" needs to look at that final input. If the CA thinks "I'm just signing my shopping list, after applying this particular transform, not a certificate" but the output of the transform happens to be a TBSCertificate, they've still signed a certificate. :-)

This is all needlessly confusing and if we had a time machine, we wouldn't have this mess of conversions and different views of the world, but here we are. :-) The upshot is:

* CA keys only generate one kind of signature, subtrees. Sometimes start is zero and sometimes it's non-zero.
* Saying "checkpoints or subtrees" costs us nothing, so we may as well say, it, even though it's actually the same thing being signed. :-)

PR #223, opened 2026-05-06
Add issuer alternative name to TBSCertificateLogEntry

FiloSottile:
Currently, CAs often have managed intermediates to let entities get a "branded" issuer. In a root + intermediate PKI hierarchy, this has little cost. In a MTC PKI, adding an issuance log is expensive (e.g. mirrors need to learn about it, clients need to be provided landmarks, etc.).

This PR provides a mechanism to attribute cosmetically certificates to "branded" issuers using a Name (for equivalence to the Issuer field) in an issuer alternative name extension, without using separate issuance logs.

I did some light profiling of the Name for implementation sanity (single-entry RDNs, UTF8String AttributeValues), one could do a lot more, but such rules are probably best left to root programs.

### DIFF
--- a/draft-ietf-plants-merkle-tree-certs.md
+++ b/draft-ietf-plants-merkle-tree-certs.md
@@ -1014,7 +1014,9 @@ The fields of a TBSCertificateLogEntry are defined as follows:
 
 * `version`, `validity`, `subject`, `issuerUniqueID`, `subjectUniqueID`, and `extensions` have the same semantics as the corresponding TBSCertificate fields, defined in {{Section 4.1.2 of !RFC5280}}.
 
-*  `issuer` is the CA ID as a PKIX distinguished name, as described in {{ca-ids}}.
+* `issuer` is the CA ID as a PKIX distinguished name, as described in {{ca-ids}}.
+
+  * The `issuer` field is not human-readable. A TBSCertificateLogEntry MAY carry a human-readable label for the CA, suitable for display in user interfaces, in an issuer alternative name extension ({{Section 4.2.1.7 of !RFC5280}}). If present, the extension MUST be marked non-critical. The `IssuerAltName` SEQUENCE MUST contain a single `GeneralName` of type `directoryName`, whose `Name` MUST use the `rdnSequence` CHOICE. Each `RelativeDistinguishedName` MUST contain a single `AttributeTypeAndValue`. The extension is purely cosmetic, and MUST NOT be used in path validation or any other trust decision. The value MUST NOT be assumed unique across issuance logs and MAY change across entries in the same issuance log.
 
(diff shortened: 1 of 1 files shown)


lukevalenta (approved):
Seems reasonable. IIUC there was nothing before this PR that restricts a CA's ability to set the `issuerAltName` extension, so this just clarifies how that extensions should be used in the MTC context.

Worth starting a PLANTS thread to see if there are other opinions, but I will go ahead and mark my approval.

davidben (approved):
Seems reasonable. Arguably this is orthogonal to MTCs, but hinting at it here SGTM.

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -972,7 +972,9 @@ The fields of a TBSCertificateLogEntry are defined as follows:
 
 * `version`, `validity`, `subject`, `issuerUniqueID`, `subjectUniqueID`, and `extensions` have the same semantics as the corresponding TBSCertificate fields, defined in {{Section 4.1.2 of !RFC5280}}.
 
-*  `issuer` is the issuance log's log ID as a PKIX distinguished name, as described in {{log-ids}}.
+* `issuer` is the issuance log's log ID as a PKIX distinguished name, as described in {{log-ids}}.
+
+  * The `issuer` field is not human-readable. A TBSCertificateLogEntry MAY carry a human-readable label for the CA, suitable for display in user interfaces, in an issuer alternative name extension ({{Section 4.2.1.7 of !RFC5280}}). If present, the extension MUST be marked non-critical. The `IssuerAltName` SEQUENCE MUST contain a single `GeneralName` of type `directoryName`, whose `Name` MUST use the `rdnSequence` CHOICE. Each `RelativeDistinguishedName` MUST contain a single `AttributeTypeAndValue`, and each `AttributeValue` MUST be a `UTF8String`. The extension is purely cosmetic, and MUST NOT be used in path validation or any other trust decision. The value MUST NOT be assumed unique across issuance logs and MAY change across entries in the same issuance log.
```
davidben: Re UTF8String, @robstradling that this breaks attributes like countryName that specifically use PrintableString, and also that RFC 5280 already has this text: https://www.rfc-editor.org/rfc/rfc5280.html#section-4.1.2.4:~:text=CAs%0A%20%20%20conforming%20to%20this%20profile%20MUST%20use%20either%20the%20PrintableString%20or%0A%20%20%20UTF8String%20encoding%20of%20DirectoryString%2C%20with%20two%20exceptions.

That's not _quite_ as strong as this text, but maybe it's good enough.

FiloSottile: That's probably good enough, yeah. Removed the part about UTF8String.

PR #226, opened 2026-05-07
Add randomizer to leafs

bwesterb:
MTC provides 128-bit security for its transparency goals: a CA needs to find a SHA-256 colission to hide a certificate from monitors.

Before this commit, an outside attacker can do something similar: find a hash colission to get a benign and malicious entry with the same hash. The attacker would then ask the CA to publish the benign entry with which it'll also get the malicious one certified. Thus forgery resistance is also capped at 128 bit.

This commit adds a CA-chosen randomizer to each leaf. This does not change transparency security: that's still 128-bit, but it does bump the forgery resistance to at least the multi-target second-preimage resistance of the hash. So that's 192+ bits with less than 2^63 leafs in the tree.

### DIFF
--- a/demo/config.go
+++ b/demo/config.go
@@ -23,6 +23,7 @@ const (
 	VersionDavidben10
 	VersionPlants01
 	VersionPlants02
+	VersionPlants04
 )
 
 func (v DraftVersion) String() string {
@@ -35,6 +36,8 @@ func (v DraftVersion) String() string {
 		return "plants-01"
 	case VersionPlants02:
 		return "plants-02"
+	case VersionPlants04:
+		return "plants-04"
 	}
 	panic(fmt.Sprintf("unknown version %d", v))
 }
@@ -62,6 +65,8 @@ func DraftVersionFromString(s string) (v DraftVersion, ok bool) {
 		return VersionPlants01, true
 	case "plants-02":
 		return VersionPlants02, true
+	case "plants-04":
+		return VersionPlants04, true
 
 	default:
 		return 0, false
--- a/demo/config.go
+++ b/demo/config.go
@@ -66,6 +66,10 @@ func DraftVersionFromString(s string) (v DraftVersion, ok bool) {
 	case "plants-02":
 		return VersionPlants02, true
 	case "plants-04":
+		// TODO: not fully up-to-date with draft-04 yet. Still missing:
+		//   - CA cosigner's ID should now be the log ID (fix in provided config file).
+		//   - Signature scheme has been updated.
+		//   - Add code to synthesize the CA certs.
 		return VersionPlants04, true
 
 	default:
--- a/draft-ietf-plants-merkle-tree-certs.md
+++ b/draft-ietf-plants-merkle-tree-certs.md
@@ -1693,13 +1693,13 @@ A key security requirement of any PKI scheme is that relying parties only accept
 
 * In landmark-relative certificates, the cosigner requirements are checked ahead of time, when the trusted subtrees are predistributed ({{trusted-subtrees}}).
 
-Given a subtree hash computed over entries that the CA certified, it must be computationally infeasible to construct an entry not on this list, and an inclusion proof, such that inclusion proof verification succeeds. A colission-resistant hash is sufficient, but not required: it's also enough for the hash to be multi-target second-preimage resistant. A secure 256-bit hash provides 256-t multi-target second-preimage resistance against 2^t targets, and thus 200-bit authenticity for a log of 2^55 entries.
+Given a subtree hash computed over entries that the CA certified, it must be computationally infeasible to construct an entry not on this list, and an inclusion proof, such that inclusion proof verification succeeds. A collision-resistant hash is sufficient, but not required: it's also enough for the hash to be multi-target second-preimage resistant. A secure 256-bit hash provides 256-t multi-target second-preimage resistance against 2^t targets, and thus 200-bit authenticity for a log of 2^55 entries.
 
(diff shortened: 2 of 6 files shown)


draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1663,18 +1668,6 @@ In particular, relying parties that share an update process for trusted subtrees
 
 # Security Considerations
 
-## Authenticity
```
davidben, line 1669: Should we replace this section with new text?

bwesterb: Moved it down so we start with transparency.

> transparency is 128-bit; forgery is 192-bit

sounds better to my ear than

> forgery is 192-bit; transparency is 128-bit

davidben: Derp, sorry, must have missed it on my phone.

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1687,6 +1680,24 @@ If the CA sends one view to some cosigners and another view to other cosigners,
...
+
+* In standalone certificates, the relying party's cosigner requirements ({{trusted-cosigners}}) are expected to include some signature by the CA's cosigner. The CA's cosigner ({{certification-authority-cosigners}}) is defined to certify the contents of every checkpoint and subtree that it signs.
+
+* In landmark-relative certificates, the cosigner requirements are checked ahead of time, when the trusted subtrees are predistributed ({{trusted-subtrees}}).
+
+Given a subtree hash computed over entries that the CA certified, it must be computationally infeasible to construct an entry not on this list, and an inclusion proof, such that inclusion proof verification succeeds. A colission-resistant hash is sufficient, but not required: it's also enough for the hash to be multi-target second-preimage resistant. A secure 256-bit hash provides 256-t multi-target second-preimage resistance against 2^t targets, and thus 200-bit authenticity for a log of 2^55 entries.
```
```suggestion
Given a subtree hash computed over entries that the CA certified, it must be computationally infeasible to construct an entry not on this list, and an inclusion proof, such that inclusion proof verification succeeds. A collision-resistant hash is sufficient, but not required: it's also enough for the hash to be multi-target second-preimage resistant. A secure 256-bit hash provides 256-t multi-target second-preimage resistance against 2^t targets, and thus 200-bit authenticity for a log of 2^55 entries.
```

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1687,6 +1680,24 @@ If the CA sends one view to some cosigners and another view to other cosigners,
...
+
+The previous assumes the attacker does not influence which entries are certified. In normal operation, an attacker can request legitimate issuance of entries. An attacker might find two entries with the same hash, and request issuance for one, leading to a certification of the second. The `randomizer` thwarts this.
+
+Log entries contain public key hashes. It must additionally be computationally infeasible to compute a public key whose hash matches an existing entry, other than the intended public key. This is again second-preimage multitarget resistance.
+
+The subject public key hash is not randomized, and so an attacker that is able to compute colissions could find two subject public keys with the same hash, which is of no security concern given the attacker needs permission to publish either anyway.
```
davidben: Same misspelling of "collision".

"needs permission to publish anyway" feels incomplete because earlier we just mentioned that attackers usually _can_ publish entries in the log. Perhaps:

```suggestion
The subject public key hash is not randomized, and so an attacker that is able to compute collisions could find two subject public keys with the same hash. However, in order to request certification of a new entry with this hash, the attacker must already be authorized for this new entry. The attacker would thus only be able to certify colliding pairs of public keys for its own identities.
```

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1687,6 +1680,24 @@ If the CA sends one view to some cosigners and another view to other cosigners,
...
+
+* In landmark-relative certificates, the cosigner requirements are checked ahead of time, when the trusted subtrees are predistributed ({{trusted-subtrees}}).
+
+Given a subtree hash computed over entries that the CA certified, it must be computationally infeasible to construct an entry not on this list, and an inclusion proof, such that inclusion proof verification succeeds. A colission-resistant hash is sufficient, but not required: it's also enough for the hash to be multi-target second-preimage resistant. A secure 256-bit hash provides 256-t multi-target second-preimage resistance against 2^t targets, and thus 200-bit authenticity for a log of 2^55 entries.
+
+The previous assumes the attacker does not influence which entries are certified. In normal operation, an attacker can request legitimate issuance of entries. An attacker might find two entries with the same hash, and request issuance for one, leading to a certification of the second. The `randomizer` thwarts this.
```
davidben: Suspect this is not immediately obvious without an example. (Also flows will into comment below...)

```suggestion
The previous assumes the attacker does not influence which entries are certified. In normal operation, an attacker can request legitimate issuance of entries, e.g. certificates for DNS names the attacker controls. An attacker might find two entries with the same hash, one of which of it is authorized to request. It might than request issuance for the authortized entry, leading to a certification of the other entry. The `randomizer` thwarts this.
```

demo/config.go
```diff
@@ -62,6 +65,8 @@ func DraftVersionFromString(s string) (v DraftVersion, ok bool) {
 		return VersionPlants01, true
 	case "plants-02":
 		return VersionPlants02, true
+	case "plants-04":
+		return VersionPlants04, true
```
davidben, line 73: This isn't *totally* up-to-date with draft-04 yet. Though yeah incrementally updating it seems good. Maybe add a TODO comment that we still need to do other bits. Things that are missing:

* CA cosigner's ID is now the log ID. I think this isn't a generator fix but a fix to the provided config file.
* Signature scheme has been updated
* Not a bug, but a missing feature: we should probably write some code to synthesize the CA certs too.

bwesterb: Added todo

davidben:
Closing this in favor of the extension block.

PR #227, opened 2026-05-07
Update for new trust-anchor-ids draft

### LINKED ISSUE #152: Be more opinionated about ID allocation

davidben:
The extension to trust-anchor-ids was moved upstream in https://github.com/tlswg/tls-trust-anchor-ids/pull/104. Update this accordingly.

As part of this, simplify landmark OIDs. The ability to have multiple landmark sequences per log is an unnecessary flexibility.  Between this and the simplified cosigner IDs, this PR closes #152. A CA need only pick one OID.

### DIFF
--- a/draft-ietf-plants-merkle-tree-certs.md
+++ b/draft-ietf-plants-merkle-tree-certs.md
@@ -907,6 +907,14 @@ Throughout this document, the hash algorithm in use is referred to as HASH, and 
 
 Each issuance log is identified by a *log ID*, which is a trust anchor ID {{!I-D.ietf-tls-trust-anchor-ids}}.
 
+When allocating log IDs, the entire object identifier (OID) arc is allocated for use with the issuance log. OIDs under this arc are allocated by Merkle Tree Certificates protocol. Given a log ID whose OID representation is `logID`, this document allocates the following OIDs:
+
+* For each non-negative integer `L`, the OID `{logID landmarks(0) L}` represents the landmark ({{landmark-tree-sizes}}) with number `L`. This OID may used as a trust anchor ID, as described in {{landmark-relative-certificates-tls}}.
+
+* For each non-negative integer `L`, the OID `{logID landmarkGroups(1) L}` represents a trust anchor group ({{Section 5 of !I-D.ietf-tls-trust-anchor-ids}}) containing landmark number `L` and earlier landmarks, as defined in {{single-log-landmark-groups}}.
+
+Future extensions to this protocol MAY define further allocations.
+
 An issuance log's log ID determines a PKIX distinguished name ({{Section 4.1.2.4 of !RFC5280}}). The distinguished name has a single relative distinguished name, which has a single attribute. The attribute has type `id-rdna-trustAnchorID`, defined below:
 
 ~~~asn.1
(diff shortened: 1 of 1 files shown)


draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -907,6 +907,14 @@ Throughout this document, the hash algorithm in use is referred to as HASH, and
 
 Each issuance log is identified by a *log ID*, which is a trust anchor ID {{!I-D.ietf-tls-trust-anchor-ids}}.
 
+When allocating log IDs, the entire object identifier (OID) arc is allocated for use with the issuance log. OIDs under this arc are allocated by Merkle Tree Certificates protocol. Given a log ID whose OID representation is `logID`, this document allocates the following OIDs:
```
```suggestion
When allocating log IDs, the entire object identifier (OID) arc is allocated for use with the issuance log. OIDs under this arc are allocated by the Merkle Tree Certificates protocol. Given a log ID whose OID representation is `logID`, this document allocates the following OIDs:
```

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1507,70 +1513,58 @@ Most X.509 fields such as subjectPublicKeyInfo and X.509 extensions such as subj
...
 
-A standalone certificate will generally be accepted by relying parties that trust the issuing CA. To determine this, a standalone certificate has a trust anchor ID of the corresponding log ID ({{log-ids}}). The authenticating party can obtain this information either by parsing the certificate's issuer field or via out-of-band information as described in {{Section 3.2 of !I-D.ietf-tls-trust-anchor-ids}}. Authenticating and relying parties SHOULD use the `trust_anchors` extension to determine whether the standalone certificate would be acceptable.
+When needed, {{Section 5 of !I-D.ietf-tls-trust-anchor-ids}} describes how PKIs requiring further size savings can use trust anchor groups that span multiple CA instances. For example, a single ID may signal support for a group of CAs from the same CA operator. This section describes how such groups can be applied to landmarks, using a variation of the versioning construction described in {{Section 5.1 of !I-D.ietf-tls-trust-anchor-ids}}.
```
lukevalenta: This should be setting up for the "one group for multiple CA operators with timestamp" case, right?
```suggestion
When needed, {{Section 5 of !I-D.ietf-tls-trust-anchor-ids}} describes how PKIs requiring further size savings can use trust anchor groups that span multiple CA instances. For example, a single ID may signal support for a group of CAs across one or more CA operators. This section describes how such groups can be applied to landmarks, using a variation of the versioning construction described in {{Section 5.1 of !I-D.ietf-tls-trust-anchor-ids}}.
```

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1507,70 +1513,58 @@ Most X.509 fields such as subjectPublicKeyInfo and X.509 extensions such as subj
...
+Trust anchor groups containing landmarks SHOULD define versions predictably based on the time. For example, if the contained CAs allocate landmarks roughly hourly, the trust anchor group might increment the version component every hour. Each given version of the group SHOULD contain the active landmarks as of the corresponding timestamp.
 
-A standalone certificate MAY also be sent without explicit relying party trust signals, however doing so means the authenticating party implicitly assumes the relying party trusts the issuing CA. This may be viable if, for example, the CA is relatively ubiquitous among supported relying parties.
+This predictable cadence allows the CA to construct trust anchor group inclusions without additional coordination. Conversely, a relying party MAY send a version if its trusted subtrees ({{trusted-subtrees}}) are up-to-date for all contained CAs, as of the versions timestamp.
```
lukevalenta: Not sure this is the best edit, but it a pointer to where trust anchor group inclusions are defined would be helpful. On my first read, it looked like it was saying that the CA is the one constructing the multi-CA-operator trust anchor group, but what's actually happening is that the CA is constructing the CertificatePropertiesList to send to the authenticating party.
```suggestion
This predictable cadence allows the CA to construct trust anchor group inclusions ({{Section 7.2 of !I-D.ietf-tls-trust-anchor-ids}}) for issued certificates without additional coordination. Conversely, a relying party MAY send a version if its trusted subtrees ({{trusted-subtrees}}) are up-to-date for all contained CAs, as of the versions timestamp.
```

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1507,70 +1513,58 @@ Most X.509 fields such as subjectPublicKeyInfo and X.509 extensions such as subj
...
+The relying party can mitigate this in a number of ways:
 
-When both a landmark and standalone certificate are supported by a relying party, an authenticating party SHOULD preferentially use the landmark-relative certificate. A landmark-relative certificate asserts the same information as its standalone counterpart, but is expected to be smaller. An authenticating party SHOULD NOT send a landmark-relative certificate without a signal that the relying party trusts the corresponding landmark subtree. Even if the relying party is assumed to trust the issuing CA, the relying party may not have sufficiently up-to-date trusted subtrees.
+* If the trust anchor group consists of CAs from the same operator, waiting until all CAs are reachable will be minimally disruptive.
+
+* The relying party can opt to send the group with an older timestamp, combined with other, smaller groups at newer timestamps to better describe its state.
```
lukevalenta, line 1567: Just a thought: should we define a mechanism in trust anchor IDs for the RP to say it has all trust anchors in a timestamp-versioned group _except_ some list of IDs from lagging CAs? That might end up being smaller on the wire than old-timestamp-plus-smaller-groups.

lukevalenta (approved):

draft-ietf-plants-merkle-tree-certs.md
davidben: I guess either, but sure. :-) I figure the decision tree for a PKI would look something like:

* Can you just list the CAs? Use the single-log groups. That's the easiest.
* Still too many? Use per-CA-operator groups. Still minimal coordination overhead.
* Still too many? Okay start making multi-CA groups. That's a bit of fuss.

draft-ietf-plants-merkle-tree-certs.md
davidben: Hehe, yeah, @bob-beck and I were talking about that. Not wrong, though it is another moving part. Could probably be added as a separate extension or a separate list in the same extension. Though we also have the retry.

(I figure that's probably separate from this PR.)

lukevalenta: Cool, yep definitely separate from this PR

lukevalenta:
Latest version LGTM!

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -907,6 +907,14 @@ Throughout this document, the hash algorithm in use is referred to as HASH, and
 
 Each issuance log is identified by a *log ID*, which is a trust anchor ID {{!I-D.ietf-tls-trust-anchor-ids}}.
 
+When allocating log IDs, the entire object identifier (OID) arc is allocated for use with the issuance log. OIDs under this arc are allocated by the Merkle Tree Certificates protocol. Given a log ID whose OID representation is `logID`, this document allocates the following OIDs:
+
+* For each non-negative integer `L`, the OID `{logID landmarks(0) L}` represents the landmark ({{landmark-tree-sizes}}) with number `L`. This OID may used as a trust anchor ID, as described in {{landmark-relative-certificates-tls}}.
+
+* For each non-negative integer `L`, the OID `{logID landmarkGroups(1) L}` represents a trust anchor group ({{Section 5 of !I-D.ietf-tls-trust-anchor-ids}}) containing landmark number `L` and earlier landmarks, as defined in {{single-log-landmark-groups}}.
```
bwesterb: Maybe add for the latter that this is what an RP would typically sent to signal which landmarks it knows about, and for the former that this is used by the AP to tell the RP which landmarks it has in the retry flow.

bwesterb (approved):

draft-ietf-plants-merkle-tree-certs.md
davidben: Good call. Done.

PR #231, opened 2026-05-15
Canonicalize the order of cosignatures in MTCProofs

### LINKED ISSUE #221: Fix the order of cosignatures?
Cosignatures are an unordered set right now. Should we insist on a particular order, like sorting the IDs lexicographically? That way a receiver doesn't need to sort to detect duplicates, and we generally limit unnecessary variation. (On the other hand, it's one more thing a CA might mess up.)

See discussion here https://github.com/ietf-plants-wg/merkle-tree-certs/pull/220#issuecomment-4376952323

(CC @ilaril)

davidben:
Closes #221.

### DIFF
--- a/draft-ietf-plants-merkle-tree-certs.md
+++ b/draft-ietf-plants-merkle-tree-certs.md
@@ -1241,7 +1241,7 @@ Each element of the `signatures` field MUST have a unique `cosigner_id`. Element
 * Shorter byte strings are ordered before longer byte strings
 * Byte strings of the same length are ordered lexicographically
 
-An MTCProof parser MUST reject the input if there are duplicate `cosigner_id` values, or if they are not ordered correctly. This can be done by checking each `cosigner_id` value comes strict after the previous one in the above order.
+An MTCProof parser MUST reject the input if there are duplicate `cosigner_id` values, or if they are not ordered correctly. This can be done by checking each `cosigner_id` value comes strictly after the previous one in the above order.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1236,6 +1236,13 @@ struct {
 
 `start` and `end` MUST contain the corresponding parameters of the chosen subtree. `inclusion_proof` MUST contain a subtree inclusion proof ({{subtree-inclusion-proofs}}) for the log entry and the subtree. `signatures` contains the chosen subtree signatures. In each signature, `cosigner_id` contains the cosigner ID ({{cosigners}}) in its binary representation ({{Section 3 of !I-D.ietf-tls-trust-anchor-ids}}), and `signature` contains the signature value as described in {{signature-format}}. The `timestamp` field used when computing the signature MUST be zero.
 
+Each element of the `signatures` field MUST have a unique `cosigner_id`. Elements MUST be ordered by `cosigner_id` as follows:
+
+* Shorter byte strings are ordered before longer byte strings
+* Byte strings of the same length are ordered lexicographically
+
+An MTCProof parser MUST reject the input if there are duplicate `cosigner_id` values, or if they are not ordered correctly. This can be done by checking each `cosigner_id` value comes strict after the previous one in the above order.
```
```suggestion
An MTCProof parser MUST reject the input if there are duplicate `cosigner_id` values, or if they are not ordered correctly. This can be done by checking each `cosigner_id` value comes strictly after the previous one in the above order.
```

lukevalenta (approved):

PR #236, opened 2026-05-18
Add an extension field to MerkleTreeCertEntry and MTCProof

cjpatton:
Define an extensions field for the MerkleTreeCertEntry structure. Repeated extensions are disallowed, but unrecognized extensions are ignored. The extensions are duplicated in the MTCProof.

This change accommodates two previously requested features for MTC:

1. https://github.com/ietf-plants-wg/merkle-tree-certs/pull/172: The accumulated count for the accumulated log entry type

2. https://github.com/ietf-plants-wg/merkle-tree-certs/pull/226: The log entry randomizer


draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1284,13 +1296,16 @@ struct {
...
 struct {
+    MerkleTreeCertEntryExtension extensions<0..2^16-1>;
     uint48 start;
     uint48 end;
     HashValue inclusion_proof<0..2^16-1>;
     MTCSignature signatures<0..2^16-1>;
 } MTCProof;
 ~~~
 
+`extensions` is the list of extensions associated with the log entry ({{log-entries}}). The extensions MUST appear in the same order here as they do in the log entry.
```
cjpatton: Reviewer note: Since the extensions are duplicated between the cert signature and the log entry, we might want to nudge implementations towards not storing the extensions in the log entry explicitly and just parse them from the signature when they need to compute the hash.

demo/mtc.json
```diff
@@ -1,5 +1,5 @@
 {
-  "Version": "plants-02",
+  "Version": "plants-04",
```
davidben: Since plants-04 isn't done yet and we don't even implement all of what's in git, maybe let's leave the default at `plants-02` for now.

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -957,11 +957,21 @@ Each entry in the log is a MerkleTreeCertEntry, defined with the TLS presentatio
 ~~~tls-presentation
 struct {} Empty;
 
+enum {
+    reserved(0), (2^16-1)
```
davidben: Probably don't even need to reserve 0, I don't think.

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1005,7 +1015,9 @@ The fields of a TBSCertificateLogEntry are defined as follows:
 
 Note the subject's public key algorithm is incorporated into both `subjectPublicKeyAlgorithm` and `subjectPublicKeyInfoHash`.
 
-MerkleTreeCertEntry is an extensible structure. Future documents may define new values for MerkleTreeCertEntryType, with corresponding semantics. See {{certification-authority-cosigners}} and {{new-log-entry-types}} for additional discussion.
+Field `extensions` is the list of tag-length-value extensions associated with the log entry. Extensions are used to override the logic for processing an entry without modifying the entry itself (e.g., by adding an extension to the TBSCertificateLogEntry). The extensions field MUST NOT contain repeated tags; the CA MUST reject entries with an extension field containing more than one instance of a given extension type. Cosigners and relying parties MUST ignore unrecognized extension types.
```
davidben: Nits:

> Extensions are used to override the logic for processing an entry without modifying the entry itself (e.g., by adding an extension to the TBSCertificateLogEntry).

If they _override_ the logic, it suggests that unrecognized extensions should be an error, but we want the RP to tolerate unknown extensions.

> the CA MUST reject entries with an extension field containing more than one instance of a given extension type.

CAs are the ones constructing these, rather than consuming these. Though checking duplicates is actually kinda annoying, unless we enforce that they're ordered numerically. (We could do that!)

Probably also something about not allowing unknown extension types. I think the existing text for entry types is in:
* Certification Authority Cosigners
* New Log Entry Types
We can probably extend those sections?

draft-ietf-plants-merkle-tree-certs.md
davidben: Which implementations? The logs don't store MTCProof structures.

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1850,6 +1866,14 @@ IANA is requested to add the following entry to the "SMI Security for PKIX Relat
 |---------|-----------------------|------------|
 | TBD     | id-rdna-trustAnchorID | [this-RFC] |
 
+## Log Entry Extensions Registry
+
+IANA is requested to create a new registry called "Merkle Tree Certificate Log Entry Extensions" with the following initial contents:
```
davidben: I suspect setting up a registry might need a bit since we have to define how new codepoints are assigned and everything. If we're not defining any codepoints yet, maybe we can get away with punting the registry to the first document that needs to do it?

draft-ietf-plants-merkle-tree-certs.md
cjpatton: ahhh yes of course :)

demo/mtc.json
cjpatton: Done.

draft-ietf-plants-merkle-tree-certs.md
cjpatton: Sure. How should we spell `MerkleTreeCertEntryExtensionType` if we don't have an initial variant?  This doesn't look syntactically correct to me:

```tls-presentation
enum {
    (2^16-1)
} MerkleTreeCertEntryExtensionType;
```

We could just do
```tls-presentation
uint16 MerkleTreeCertEntryExtensionType;
```

draft-ietf-plants-merkle-tree-certs.md
cjpatton: > If they _override_ the logic, it suggests that unrecognized extensions should be an error, but we want the RP to tolerate unknown extensions.

Agreed. How would you write this instead? Hows about "extends the logic with optional behavior"?

> CAs are the ones constructing these, rather than consuming these. Though checking duplicates is actually kinda annoying, unless we enforce that they're ordered numerically. (We could do that!)

Ah, you mean that the CA would add the extensions instead of the party requesting the cert? That makes sense.

I'd be fine with allowing duplicates, but what should we tell RPs to do when they encounter them? Should we leave it up to the extension to specify their behavior?

> Probably also something about not allowing unknown extension types.

Do you mean the CA shouldn't add extensions to an entry of a type it doesn't recognize?

draft-ietf-plants-merkle-tree-certs.md
cjpatton: Done.

draft-ietf-plants-merkle-tree-certs.md
davidben: It is quite unusual, but I think `enum { (2^16-1) } MerkleTreeCertEntryExtensionType` is fine? I don't know if I've ever seen it used, but eh. The TLS presentation language isn't really a formal language that anyone tries to parse, so let's ask for forgiveness later if anyone thinks this is wrong. :-)

draft-ietf-plants-merkle-tree-certs.md
davidben: > Agreed. How would you write this instead? Hows about "extends the logic with optional behavior"?

This is a cop-out, but we could just not say anything. They're additional values associated with the entry, but the extension has to define the semantics. We can even explicitly say something like "This document does not define any extensions, but future documents MAY define extensions."

> Ah, you mean that the CA would add the extensions instead of the party requesting the cert? That makes sense.

Yeah. For both examples we have (randomizer and accumulator count), it's the CA constructing this as well.

I mean, maybe someone will define an extension that includes some value that the subscriber requests, but I think that still counts as the CA adding the extension, even if it's decided, for this one field, it's supposed to copy the value from the certificate request. (But I doubt that'll happen.)

> I'd be fine with allowing duplicates, but what should we tell RPs to do when they encounter them? Should we leave it up to the extension to specify their behavior?

Oh, I don't think we should allow duplicates, so we don't have to think about it. It's just that detecting duplications is mildly annoying if you have to copy the list somewhere and then sort, rather than just iterating through the data in-place. I've seen TLS stacks that look at the first or last value if there's a duplicate, and I've seen ones that only detect duplicates of known extensions and don't notice duplicates of unknown ones.

Mirroring #231, I'm inclined to say they have to be ordered numerically, which makes checking for duplicates easy. WDYT?

> Do you mean the CA shouldn't add extensions to an entry of a type it doesn't recognize?

Yeah. Perhaps update these sections:
* https://ietf-plants-wg.github.io/merkle-tree-certs/draft-ietf-plants-merkle-tree-certs.html#name-log-entries:~:text=MerkleTreeCertEntry%20is%20an%20extensible%20structure.%20Future%20documents%20may%20define%20new%20values%20for%20MerkleTreeCertEntryType%2C%20with%20corresponding%20semantics.%20See%20Section%205.4%20and%20Section%2012.5%20for%20additional%20discussion.
* https://ietf-plants-wg.github.io/merkle-tree-certs/draft-ietf-plants-merkle-tree-certs.html#section-5.4-5:~:text=Entries%20are%20extensible,of%20new%20formats.
* https://ietf-plants-wg.github.io/merkle-tree-certs/draft-ietf-plants-merkle-tree-certs.html#name-new-log-entry-types (rename to Log Entry Extensions?)

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1459,6 +1474,7 @@ This procedure only replaces the signature verification portion of X.509 path va
 In this procedure, `entry_hash` can equivalently be computed in a single pass from the DER-encoded TBSCertificate, without storing the full TBSCertificateLogEntry or MerkleTreeCertEntry in memory:
 
 1. Initialize a hash instance.
+1. Write the `extensions` field from the MTCProof to the hash.
```
davidben, line 1480: Oh, the step that begins "Construct a MerkleTreeCertEntry of type `tbs_cert_entry`" above also needs to be updated. (GitHub doesn't let me anchor a review comment to text you didn't change, unhelpfully.)

draft-ietf-plants-merkle-tree-certs.md
cjpatton: Done!

draft-ietf-plants-merkle-tree-certs.md
cjpatton: > > Agreed. How would you write this instead? Hows about "extends the logic with optional behavior"?
> 
> This is a cop-out, but we could just not say anything. They're additional values associated with the entry, but the extension has to define the semantics. We can even explicitly say something like "This document does not define any extensions, but future documents MAY define extensions."

Done. Trimmed to just "`extensions` is the list of tag-length-value extensions associated with the log entry." and moved this paragraph to follow the tls-presentation block defining `MerkleTreeLogEntry`.


> > I'd be fine with allowing duplicates, but what should we tell RPs to do when they encounter them? Should we leave it up to the extension to specify their behavior?
> 
> Oh, I don't think we should allow duplicates, so we don't have to think about it. It's just that detecting duplications is mildly annoying if you have to copy the list somewhere and then sort, rather than just iterating through the data in-place. I've seen TLS stacks that look at the first or last value if there's a duplicate, and I've seen ones that only detect duplicates of known extensions and don't notice duplicates of unknown ones.
> 
> Mirroring [#231](https://github.com/ietf-plants-wg/merkle-tree-certs/pull/231), I'm inclined to say they have to be ordered numerically, which makes checking for duplicates easy. WDYT?

Works for me, done. Added MUST be sorted and MUST NOT contain duplicates. Removed text about enforcement, since here we're specifying how to construct entries.

> > Do you mean the CA shouldn't add extensions to an entry of a type it doesn't recognize?
> 
> Yeah. Perhaps update these sections:
> 
>     * https://ietf-plants-wg.github.io/merkle-tree-certs/draft-ietf-plants-merkle-tree-certs.html#name-log-entries:~:text=MerkleTreeCertEntry%20is%20an%20extensible%20structure.%20Future%20documents%20may%20define%20new%20values%20for%20MerkleTreeCertEntryType%2C%20with%20corresponding%20semantics.%20See%20Section%205.4%20and%20Section%2012.5%20for%20additional%20discussion.

Done.

>     * https://ietf-plants-wg.github.io/merkle-tree-certs/draft-ietf-plants-merkle-tree-certs.html#section-5.4-5:~:text=Entries%20are%20extensible,of%20new%20formats.

Done.

>     * https://ietf-plants-wg.github.io/merkle-tree-certs/draft-ietf-plants-merkle-tree-certs.html#name-new-log-entry-types (rename to Log Entry Extensions?)

Question here: Should an RP accept an entry with unrecognized extensions? It seems like the intent of both use cases (randomizer and accumlator) is that the RP can safely ignore these. In other words, I don't think we actually want any changes here?

draft-ietf-plants-merkle-tree-certs.md
cjpatton: Good catch, done.

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -971,6 +979,8 @@ struct {
 } MerkleTreeCertEntry;
 ~~~
 
+Field `extensions` is the list of tag-length-value extensions associated with the log entry. The extensions list MUST be appear in ascending order by tag and MUST NOT contain more than one instance of the same tag.
```
davidben: Maybe this so it's clear what we mean by "tag".

```suggestion
Field `extensions` is the list of tag-length-value extensions associated with the log entry. The extensions list MUST be appear in ascending order by `extension_type` and MUST NOT contain two extensions with the same `extension_type`.
```

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1176,7 +1186,7 @@ What it means to certify an entry depends on the entry type:
 * To certify an entry of type `null_entry` is a no-op. A CA MAY freely certify `null_entry` without being held responsible for any validation.
 * To certify an entry of type `tbs_cert_entry` is to certify the TBSCertificateLogEntry, as defined in {{log-entries}}.
 
-Entries are extensible. Future documents MAY define `type` values and what it means to certify them. A CA MUST NOT sign a subtree if it contains an entry with `type` that it does not recognize. Doing so would certify that the CA has validated the information in some not-yet-defined entry format. {{new-log-entry-types}} further discusses security implications of new formats.
+Entries are extensible. Future documents MAY define `type` values and what it means to certify them. Likewise for extensions. A CA MUST NOT sign a subtree if it contains an entry with `type` or extension that it does not recognize. Doing so would certify that the CA has validated the information in some not-yet-defined entry format. {{new-log-entry-types}} further discusses security implications of new formats.
```
```suggestion
Entries are extensible. Future documents MAY define `type` and `extension_type` values and what it means to certify them. A CA MUST NOT sign a subtree if it contains an entry with `type` or `extension_type` that it does not recognize. Doing so would certify that the CA has validated the information in some not-yet-defined format. {{new-log-entry-types}} further discusses security implications of such extensions.
```

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1284,13 +1294,16 @@ struct {
...
 struct {
+    MerkleTreeCertEntryExtension extensions<0..2^16-1>;
     uint48 start;
     uint48 end;
     HashValue inclusion_proof<0..2^16-1>;
     MTCSignature signatures<0..2^16-1>;
 } MTCProof;
 ~~~
 
+`extensions` is the list of extensions associated with the log entry ({{log-entries}}). The extensions MUST appear in the same order here as they do in the log entry.
```
davidben: I guess we've already constrained the order, so we can probably just say.

```suggestion
`extensions` MUST contain the log entry's `extensions` value ({{log-entries}}).
```

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1438,7 +1451,7 @@ When verifying the signature of an X.509 certificate (Step (a)(1) of {{Section 6
    1. Set `subjectPublicKeyAlgorithm` to the `algorithm` field of the `subjectPublicKeyInfo`.
    1. Set `subjectPublicKeyInfoHash` to the hash of the DER encoding of `subjectPublicKeyInfo`.
 
-1. Construct a MerkleTreeCertEntry of type `tbs_cert_entry` with contents the TBSCertificateLogEntry. Let `entry_hash` be the hash of the entry, `MTH({entry}) = HASH(0x00 || entry)`, as defined in {{Section 2.1.1 of !RFC9162}}.
+1. Construct a MerkleTreeCertEntry of type `tbs_cert_entry` with contents the TBSCertificateLogEntry and the `extensions` field of the MTCProof. Let `entry_hash` be the hash of the entry, `MTH({entry}) = HASH(0x00 || entry)`, as defined in {{Section 2.1.1 of !RFC9162}}.
```
davidben: Nit: Now that we have two fields, maybe "contents" is ambiguous. Also I see that saying the contents are the TBSCertificateLogEntry is a bit ambiguous because we encode it funny. How about...

```suggestion
1. Construct a MerkleTreeCertEntry as follows:
   1. Set `type` to `tbs_cert_entry`.
   2. Set `extensions` to the MTCProof's `extensions` value.
   3. Set `tbs_cert_entry_data` to the TBSCertificateLogEntry, encoded as described in {{log-entries}}.
   
1. Let `entry_hash` be the hash of the entry, `MTH({entry}) = HASH(0x00 || entry)`, as defined in {{Section 2.1.1 of !RFC9162}}.
```

draft-ietf-plants-merkle-tree-certs.md
davidben: I think we need to rename and revise New Log Entry Types. (If you prefer, feel free to leave a TODO and I can take a pass. This PR has already been very helpful!)

davidben (approved):

draft-ietf-plants-merkle-tree-certs.md
davidben: > Question here: Should an RP accept an entry with unrecognized extensions? It seems like the intent of both use cases (randomizer and accumlator) is that the RP can safely ignore these.

That's right.

> In other words, I don't think we actually want any changes here [in https://ietf-plants-wg.github.io/merkle-tree-certs/draft-ietf-plants-merkle-tree-certs.html#name-new-log-entry-types]?

Oh, I just meant tweaking the text a bit to talk about it at all.

draft-ietf-plants-merkle-tree-certs.md
cjpatton: Up to you, I'm happy to take a stab at it first.

I don't have the merge button. Feel free to merge whenever.

davidben (approved):

PR #239, opened 2026-05-23
Catch demo's plants-04 mode up to GitHub

davidben:
I think this is everything? Only thing that's missing is we should write out the CA too. I made the changes in separate commits, so recommend looking at commits individually.


demo/main.go
```diff
@@ -250,12 +253,12 @@ func do() error {
...
 	}
 	var signedNote bytes.Buffer
-	fmt.Fprintf(&signedNote, "%s\n", tlogOrigin(config.LogID))
+	fmt.Fprintf(&signedNote, "%s\n", tlogOrigin(LogID(&config)))
 	fmt.Fprintf(&signedNote, "%d\n", len(entries))
 	fmt.Fprintf(&signedNote, "%s\n\n", base64.StdEncoding.EncodeToString(checkpointHash[:]))
 	for i := range config.Cosigners {
 		cosigner := &config.Cosigners[i]
-		cosig, err := Cosign(cosigner, config.LogID, 0, len(entries), &checkpointHash)
+		cosig, err := Cosign(config.Version, cosigner, config.ID, 0, len(entries), &checkpointHash)
```
lukevalenta: Should be log ID, not CA ID here (matches https://github.com/ietf-plants-wg/merkle-tree-certs/pull/239/changes#diff-675ee652d306e2f15db1b2424699d9c0de5ae55b9a86d3c954c349ab8748580fR326):
```suggestion
		cosig, err := Cosign(config.Version, cosigner, LogID(&config), 0, len(entries), &checkpointHash)
```

demo/main.go
```diff
@@ -77,7 +73,14 @@ func do() error {
 	}
 
 	// Entries in the issuance log.
-	entries := [][]byte{MarshalNullEntry(config.Version)}
+	var entries [][]byte
+	if config.Version < VersionPlants04 {
+		// Prior to plants-04, the log began with a null entry.
+		entries = append(entries, MarshalNullEntry(config.Version))
+	} else if config.LogNumber == 0 {
+		// Starting plants-04, non-zero serials come from the log number.
```
lukevalenta, line 97: Ah I hadn't realized that log numbers removed the need for the initial entry. Nice. Currently the spec doesn't give any reason for `null_entry` to exist. No harm in keeping it, but maybe we could list some potential uses for it.

davidben:
Updated to also output a CA certificate, which I think does the last of it?

demo/main.go
davidben: Ah yeah. Filed https://github.com/ietf-plants-wg/merkle-tree-certs/issues/241

lukevalenta (approved):

PR #244, opened 2026-05-29
Take another pass at describing subtrees

davidben:
Punt all unused properties into the explanations section (full vs partial, directly contained), and add a section explaining what the numerical property is about.

We still mention full/partial in the explanation section, just because the subtree consistency proof evaluation talks about it. To that end, replace "full tree" with "original tree" when we're talking about the big tree that you used to pull a subtree out.

### DIFF
--- a/draft-ietf-plants-merkle-tree-certs.md
+++ b/draft-ietf-plants-merkle-tree-certs.md
@@ -430,15 +430,13 @@ A *subtree* of this Merkle Tree is itself a Merkle Tree, defined by `MTH(D[start
 *  `0 <= start < end <= n`
 * `start` is a multiple of `BIT_CEIL(end - start)`
 
-Note that, if `start` is zero, the second condition is always true.
+The second condition ensures that `MTH(D[start:end])`, built over `D[start:end]` as an independent list, is sufficiently aligned with the original Merkle Tree to support subtree consistency proofs. See {{subtrees-explain}} for more details. Note that, if `start` is zero, this second condition is always true.
 
-In the context of a single Merkle Tree, the subtree defined by `start` and `end` is denoted by half-open interval `[start, end)`. It contains the entries whose indices are in that half-open interval.
+The *size* of the subtree is `end - start`.
 
-The *size* of the subtree is `end - start`. If the subtree's size is a power of two, it is said to be *full*, otherwise it is said to be *partial*.
+In the context of a single Merkle Tree, this document denotes subtree `MTH(D[start:end])` by half-open interval `[start, end)`. It contains the entries whose indices are in that half-open interval.
 
-If a subtree is full, then it is directly contained in the tree of hash operations in `MTH(D_n)` for `n >= end`.
-
-If a subtree is partial, it is directly contained in `MTH(D_n)` only if `n = end`.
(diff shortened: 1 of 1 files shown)


draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -2048,6 +2054,58 @@ In a tree of the next power of two size, the skipped nodes in this path are wher
...
+    |[1,3)| |[3,5)|              level 1
+    +-----+ +-----+
+      / \     / \
+    +-+ +-+ +-+ +-+
+    |1| |2| |3| |4|              level 0
+    +-+ +-+ +-+ +-+
+~~~
+{: #fig-misaligned-tree title="An example misaligned tree"}
+
+The numerical constraints on `start` and `end` in {{definition-of-a-subtree}} restrict subtrees to avoid this. A Merkle Tree built over `[start, end)` has size `end - start`, and is constructed as if `start` were the first element of the sequence at index zero. To be aligned, `start` must be the leftmost leaf of the lowest common ancestor of `start` and `end` in the original tree:
```
lukevalenta: First edit is just to make it clear what "avoid this" refers to, since there's a diagram between this paragraph and the previous one (but feel free to keep as-is).

Second edit is because `end - 1` is the last leaf node in the tree.

```suggestion
The numerical constraints on `start` and `end` in {{definition-of-a-subtree}} restrict subtrees to ensure that they are properly aligned with the original tree as to permit subtree consistency proofs. A Merkle Tree built over `[start, end)` has size `end - start`, and is constructed as if `start` were the first element of the sequence at index zero. To be aligned, `start` must be the leftmost leaf of the lowest common ancestor of `start` and `end - 1` in the original tree:
```

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -2048,6 +2054,58 @@ In a tree of the next power of two size, the skipped nodes in this path are wher
...
+
+* Numerically, this means the least significant `BIT_WIDTH(end - start - 1)` bits of `start` must be zero. Equivalently, `start` must be divisible by `BIT_CEIL(end - start)`.
+
+* In the tree, this means subtrees are constructed by taking any node in the tree, setting `start` to the leftmost leaf under the node, and `end` to one past any other leaf under the node.
+
+Though most nodes overlap, not every node of the subtree is necessarily in the larger Merkle Tree, as shown in {{fig-subtree-containment-example-2}}. In general:
+
+* Subtrees whose sizes are a power of two are called *full subtrees*. A full subtree's root node will always be in the original tree.
+
+* Subtree whose sizes are not a power of two are called *partial subtrees*. A partial subtree's root node will be in the original tree of size `n`, if and only if `n = end`. Otherwise, non-leaf nodes along the partial subtree's right edge will not be part of the original tree.
```
```suggestion
* Subtrees whose sizes are not a power of two are called *partial subtrees*. A partial subtree's root node will be in the original tree of size `n`, if and only if `n = end`. Otherwise, non-leaf nodes along the partial subtree's right edge will not be part of the original tree.
```

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -2048,6 +2054,58 @@ In a tree of the next power of two size, the skipped nodes in this path are wher
...
+
+The difference between full and partial subtrees do not impact their usage, but they can help in understanding the proof constructions below.
```
```suggestion
The difference between full and partial subtrees does not impact their usage, but they can help in understanding the proof constructions below.
```

lukevalenta (approved):
Looks good!

PR #249, opened 2026-06-03
Discuss subordinate CAs and transparency

### LINKED ISSUE #146: What happens when a MTC CA issues an unconstrained intermediate certificate?
When we recast MTC as X.509, we picked up all of X.509's semantics, including the fact that a CA can, at least mechanically, issue non-MTC intermediate certificates. Doing so partially bypasses transparency benefits because only the intermediate is logged, not certificates chaining off of it.

I say partially because the fact that the CA has done so is definitely visible in the log. But if the intermediate is unconstrained, and a PKI is okay with it, then we don't actually know what certificates are valid. But this might vary a lot by PKI:

* Maybe your PKI is okay with transparency stopping at an intermediate, as long as it has name constraints, but it's not okay with intermediates without name constraints.

* Maybe your PKI is the above, but its notion of "name" is something other than SANs + name constraints and is looking for a random other X.509 extension

* Maybe you do not want to support this at all and want to forbid all intermediates

* Maybe you're just unabashedly okay with all transparency stopping at any intermediate for... some reason? I can't come up with a reason for this.

If there's a range of options, one plausible, if a bit unsatisfying, answer is that we solve this by policy. X.509 often forces us down this path, sadly. :-( We say that the MTC mechanism itself has no opinion on this, but that a relying party and root program might (and probably should!) opine on this and say what they do and don't allow.

In that case, the spec should say something about this and note that you need to pick something. We also need the technical mechanisms to apply that policy.

At a bare minimum, the technical mechanism is "monitors can see everything the CA signs and know to look for this" and "relying parties will hold the CA responsible for detected violations in trust decisions". If you see a CA=true certificate in the log, you need to treat it as impacting _every_ name (up to constraints) and monitor it it accordingly.

You might also want relying party enforcement in code so that, even if it is in there, the relying party won't accept it. For at least the "this CA should not issue intermediates period" policy, there is _almost_ a natural mechanism in X.509 already: path length constraints.

But only _almost_ because X.509 special-cases self-issued certificates everywhere as part of key rollover. Is it worth updating RFC 5280 to knock out the self-issued special case? (Alternatively, the enforcement can just be using some extra mechanism. Relying parties add checks atop RFC 5280 path validation all the time.)

davidben:
Closes #146 by opting to just describe the issue and recommending a policy- and monitoring-level fix. That's probably the simplest outcome given X.509.

### DIFF
--- a/draft-ietf-plants-merkle-tree-certs.md
+++ b/draft-ietf-plants-merkle-tree-certs.md
@@ -1855,7 +1855,7 @@ Domain separation analysis based on the structures themselves is fragile, partic
 
 Merkle Tree Certificates' transparency properties only apply to certificates directly issued by the CA, not certification paths. The CA might issue a certificate that describes an unconstrained, subordinate, non-MTC CA. Certificates issued by the subordinate CA would not be visible in the MTC CA's issuance log and thus may not be visible to monitors. However, the certificate that enables this transparency bypass will still be visible in the issuance logs.
 
-Relying parties with transparency requirements SHOULD impose policies that prevent these bypasses. For example, a relying party might require that all subordinate CAs have name constraints ({{Section 4.2.1.10 of !RFC5280}}) or forbid subordinate CAs entirely. Relying parties MAY enforce these policies as part of certification path validation.
+Relying parties with transparency requirements SHOULD impose policies that prevent these bypasses. For example, a relying party might require that all subordinate CAs have name constraints ({{Section 4.2.1.10 of !RFC5280}}) or forbid subordinate CAs entirely. Relying parties SHOULD enforce these policies as part of certification path validation.
 
 Monitors SHOULD monitor for adherence to applicable policies as part of monitoring for unauthorized certificates. For example, a monitor that looks for entries covering `example.com` SHOULD look for either a subject alternative name ({{Section 4.2.1.6 of !RFC5280}}) of `example.com` or a basic constraints ({{Section 4.2.1.9 of !RFC5280}}) extension with the cA boolean set to true.
 
--- a/draft-ietf-plants-merkle-tree-certs.md
+++ b/draft-ietf-plants-merkle-tree-certs.md
@@ -2440,4 +2440,4 @@ In draft-04, there is no fast issuance mode. In draft-05, frequent, non-landmark
 
 - Editorial fixes
 
-- Discuss the implications of subordinate CAs in Security Consideratiosn
+- Discuss the implications of subordinate CAs in Security Considerations 
--- a/draft-ietf-plants-merkle-tree-certs.md
+++ b/draft-ietf-plants-merkle-tree-certs.md
@@ -1853,9 +1853,11 @@ Domain separation analysis based on the structures themselves is fragile, partic
 
 ## Subordinate Certification Authorities
 
-Merkle Tree Certificates' transparency properties only apply to certificates directly issued by the CA, not certification paths. The CA might issue a certificate that describes an unconstrained, subordinate, non-MTC CA. Certificates issued by the subordinate CA would not be visible in the MTC CA's issuance log and thus may not be visible to monitors. However, the certificate that enables this transparency bypass will still be visible in the issuance logs.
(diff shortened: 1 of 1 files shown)


draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1851,6 +1851,16 @@ These ASN.1 structures all begin with a SEQUENCE tag, which is encoded in DER as
...
+
+Merkle Tree Certificates' transparency properties only apply to certificates directly issued by the CA, not certification paths. The CA might issue a certificate that describes an unconstrained, subordinate, non-MTC CA. Certificates issued by the subordinate CA would not be visible in the MTC CA's issuance log and thus may not be visible to monitors. However, the certificate that enables this transparency bypass will still be visible in the issuance logs.
+
+Relying parties with transparency requirements SHOULD impose policies that prevent these bypasses. For example, a relying party might require that all subordinate CAs have name constraints ({{Section 4.2.1.10 of !RFC5280}}) or forbid subordinate CAs entirely. Relying parties MAY enforce these policies as part of certification path validation.
+
+Monitors SHOULD monitor for adherence to this policy as part of monitoring for unauthorized certificates. For example, a monitor that looks for entries covering `example.com` SHOULD look for either a subject alternative name ({{Section 4.2.1.6 of !RFC5280}}) of `example.com` or a basic constraints ({{Section 4.2.1.9 of !RFC5280}}) extension with the cA boolean set to true.
```
robstradling: Which policy is "this policy"?  The context from the previous paragraph is "policies" (plural) that might not exist.

Suggestion:
s/this policy/applicable policies/

davidben: Good point. Done.

lukevalenta (approved):

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1851,6 +1851,16 @@ These ASN.1 structures all begin with a SEQUENCE tag, which is encoded in DER as
...
 
+## Subordinate Certification Authorities
+
+Merkle Tree Certificates' transparency properties only apply to certificates directly issued by the CA, not certification paths. The CA might issue a certificate that describes an unconstrained, subordinate, non-MTC CA. Certificates issued by the subordinate CA would not be visible in the MTC CA's issuance log and thus may not be visible to monitors. However, the certificate that enables this transparency bypass will still be visible in the issuance logs.
+
+Relying parties with transparency requirements SHOULD impose policies that prevent these bypasses. For example, a relying party might require that all subordinate CAs have name constraints ({{Section 4.2.1.10 of !RFC5280}}) or forbid subordinate CAs entirely. Relying parties MAY enforce these policies as part of certification path validation.
```
aarongable: I think the stronger SHOULD here is appropriate; yes there are some policy decisions which cannot be enforced at validation time, but that should be the exception, not the rule.
```suggestion
Relying parties with transparency requirements SHOULD impose policies that prevent these bypasses. For example, a relying party might require that all subordinate CAs have name constraints ({{Section 4.2.1.10 of !RFC5280}}) or forbid subordinate CAs entirely. Relying parties SHOULD enforce these policies as part of certification path validation.
```

aarongable (approved):
LGTM with one suggestion

draft-ietf-plants-merkle-tree-certs.md
davidben: Fair enough. I put MAY because I imagine, say, a command-line tool with basic verification would not enforce this. But then you could argue such a tool is probably not checking transparency and thus doesn't have transparency requirements.

bwesterb (approved):

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -2420,3 +2432,12 @@ In draft-04, there is no fast issuance mode. In draft-05, frequent, non-landmark
...
 - Add an extensions block to log entries
+
+## Since draft-ietf-plants-merkle-tree-certs-04
+{:numbered="false"}
+
+- Fix some mistakes in the single-pass signature verification algorithm
+
+- Editorial fixes
+
+- Discuss the implications of subordinate CAs in Security Consideratiosn
```
davidben: Self-review, typo:

```suggestion
- Discuss the implications of subordinate CAs in Security Considerations 
```

PR #261, opened 2026-06-25
Add a maxSerial field to the CA format

### LINKED ISSUE #260: No explicit bound on issuance log size (index/tree_size)
Section "Issuance Logs" describes an issuance log as an append-only sequence of entries identified by an index starting from zero, 
with no stated upper bound. The only explicitly bounded quantity in this area is the log number itself:

> "Log numbers are numbered consecutively from 1 to at most 65535 (2^16-1)."

However, the certificate format depends on the log entry index and subtree boundaries fitting in 48 bits.
Section "Certificate Format" defines:

> "The `serialNumber` MUST be equal to `(log_number << 48) | index`."

and `MTCProof`'s `start` and `end` fields are typed as `uint48`. Verification reconstructs the split the same way:

> "Let `index` be the least significant 48 bits of `serial` and let `log_number` be `serial >> 48`."

None of this is tied to a normative bound on `index` or the log's `tree_size`. 
There does not appear to be a statement requiring an issuance log to remain small enough for all certificate-relevant values (`index`, `start`, and `end`) to fit in `uint48`.

If a log grew beyond the range representable by the certificate fields, the certificate format would no longer have a well-defined way to represent later entry indexes in `serialNumber` or subtree boundaries in `MTCProof.start` / `MTCProof.end`. 
This is the same kind of asymmetry that is already handled for log numbers, which are explicitly bounded at 65535; `index` / `tree_size` do not seem to have the corresponding statement.

**Suggested fix:** Add an explicit requirement defining the maximum allowed issuance log tree size, chosen so that all certificate-relevant indexes and subtree boundaries fit in `uint48`.

davidben:
I suspect the discussion around revoking, log numbers, min/max indices, the CA format, etc., can be refactored given how scattered it has become. Ah well.

(Also closes #260)

### DIFF
--- a/draft-ietf-plants-merkle-tree-certs.md
+++ b/draft-ietf-plants-merkle-tree-certs.md
@@ -958,7 +958,7 @@ Each issuance log has a *log ID*, which is a trust anchor ID constructed by conc
 
 A log ID specifies both the CA and the log number in a single ID.
 
-Each issuance log describes an append-only sequence of *entries* ({{log-entries}}). Each entry is identified consecutively by an integer *index* value, starting from zero. Indices are at most 2<sup>48</sup>-1. Each entry is an assertion that the CA has certified. The entries in the issuance log are represented as a Merkle Tree, described in {{Section 2.1 of !RFC9162}}.
+Each issuance log describes an append-only sequence of *entries* ({{log-entries}}). Each entry is identified by an integer *index*, assigned consecutively starting from zero. Indices are at most 2<sup>48</sup>-1. Each entry is an assertion that the CA has certified. The entries in the issuance log are represented as a Merkle Tree, described in {{Section 2.1 of !RFC9162}}.
 
 Each log additionally maintains a *minimum index* value, which is the index of the first log entry which is available. See {{log-pruning}}. This value changes over the lifetime of the log.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -958,7 +958,7 @@ Each issuance log has a *log ID*, which is a trust anchor ID constructed by conc
 
 A log ID specifies both the CA and the log number in a single ID.
 
-Each issuance log describes an append-only sequence of *entries* ({{log-entries}}), identified consecutively by an index value, starting from zero. Each entry is an assertion that the CA has certified. The entries in the issuance log are represented as a Merkle Tree, described in {{Section 2.1 of !RFC9162}}.
+Each issuance log describes an append-only sequence of *entries* ({{log-entries}}). Each entry is identified consecutively by an integer *index* value, starting from zero. Indices are at most 2<sup>48</sup>-1. Each entry is an assertion that the CA has certified. The entries in the issuance log are represented as a Merkle Tree, described in {{Section 2.1 of !RFC9162}}.
```
lukevalenta: This reads slightly better to me, but either is fine
```suggestion
Each issuance log describes an append-only sequence of *entries* ({{log-entries}}). Each entry is identified by an integer *index*, assigned consecutively starting from zero. Indices are at most 2<sup>48</sup>-1. Each entry is an assertion that the CA has certified. The entries in the issuance log are represented as a Merkle Tree, described in {{Section 2.1 of !RFC9162}}.
```

lukevalenta (approved):

PR #264, opened 2026-07-01
Repaint the link relation bikeshed

davidben:
Based on feedback from mnot


davidben:
@aarongable LGTY?

davidben:
(Ah whoops, misremembered that your suggestion was `acme-optional-alternate`, not `acme-optional-cert`. Switched to that.)

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1672,25 +1672,21 @@ The relying party can mitigate this in a number of ways:
...
-
-ACME clients can fetch enhancement URLs to collect additional alternate certificate chains. If the resource is unavailable, the ACME client SHOULD NOT fail the overall transaction. If the resource returns an HTTP 202 (Accepted) response, the ACME client SHOULD retry the request later, incorporating any Retry-After header, but it SHOULD NOT block deployment of other chains on this process.
+An ACME client MAY fetch these URLs to collect additional alternate certificate chains. If the resource is unavailable, the ACME client SHOULD NOT fail the overall transaction. If the resource returns an HTTP 202 (Accepted) response, the ACME client SHOULD retry the request later, incorporating any Retry-After header, but it SHOULD NOT block deployment of other chains on this process.
```
jdeblasio: nit: I feel slightly weird about saying "SHOULD NOT block deployment of other chains". I think a rollout scenario we need not discourage is "renew early; hold off on rolling out new certs until landmarks are ready"

davidben: Hmm. I guess there's a whole spectrum of models here. Strictly speaking, this text is talking about deploying other changes, not un-deploying your old ones. I'll give a stab at rephrasing this, but the problem is there's a whole spectrum of different strategies, that depend on whether you currently have certificates or if your old one is expiring. It very quickly becomes too complicated to understand.

davidben: Rephrased.

jdeblasio: Thank you! That seems like a reasonable threading of the needle to me.

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1672,25 +1672,23 @@ The relying party can mitigate this in a number of ways:
...
+This behavior is not ideal for a landmark-relative certificate, which is available asynchronously and is not intended to delay the corresponding standalone certificate. This section defines the "acme-optional-alternate" link relation. When serving a certificate, an ACME server MAY provide one or more link relation header fields of type "acme-optional-alternate". "acme-optional-alternate" identifies an alternate certificate chain, but one that is optional. Relying parties that accept the optional alternate are expected to also accept either the original certificate chain or chains served under the "alternate" link relation. If the certificate chain is not yet available, the "acme-optional-alternate" URL MAY serve an HTTP 202 (Accepted) response, with a Retry-After header ({{Section 10.2.3 of !RFC9110}}) estimating when it will become available.
```
aarongable: I'd strengthen the MAY at the end of this paragraph to a SHOULD, but that opinion is weakly held.

_edit:_ Wait, I see that the paragraph in the next section says SHOULD. I think this document should probably only make statements about the HTTP 202 and Retry-After header in one place, to avoid contradicting itself.

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1672,25 +1672,23 @@ The relying party can mitigate this in a number of ways:
...
 
-This behavior is not ideal for landmark-relative certificates, which are available asynchronously and should not block deployment of their corresponding standalone certificate. The "enhancement" link relation allows an ACME server to specify which chains are necessary to fulfill the ACME order and which are optional additions.
+An ACME client MAY fetch these URLs to collect additional alternate certificate chains. If the resource is unavailable, the ACME client SHOULD NOT fail the overall transaction. If the resource returns an HTTP 202 (Accepted) response, the ACME client SHOULD retry the request later, incorporating any Retry-After header, but this process SHOULD be independent of deploying other chains in the ACME order. In particular, if deploying a new service, the ACME client SHOULD NOT block deployment on optional alternates.
```
aarongable: nit: I'm not sure that "incorporating" is the right word here (it sounds like the client should include the Retry-After header in its own follow-up request, which is obviously not the right thing to do). Maybe just `s/incorporating/according to/`?

draft-ietf-plants-merkle-tree-certs.md
davidben: Done. MAY strengthened to SHOULD and then reworked the second reference to just cite this one. (The two places are an awkward attempt to define acme-optional-alternate as a general ACME feature, but then the MTC section needs to tell you to use this feature.)

draft-ietf-plants-merkle-tree-certs.md
davidben: Done

aarongable (approved):

PR #270, opened 2026-07-14
Add special landmark group that expresses just timestamp

bwesterb:


draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1670,6 +1670,14 @@ The relying party can mitigate this in a number of ways:
 
 * A client relying party can send the newer timestamp and, in the event the unreachable CA did issue recent landmark-relative certificates, rely on the retry mechanism described in {{Section 4.3 of !I-D.ietf-tls-trust-anchor-ids}} to recover from any signaling failures.
 
+#### Universal Timestamped Landmark Group
+
+For the case where a relying party wants to express the active landmarks it has available, without detailing which CAs in particular, we provide the *universal timestamped landmark group* as follows.
+
+* The identifying OID arc of the group is `id-mtc-timestamp` (`44363.47.3`).
```
cjpatton, line 1677: As in other places where we use a `44353.47.*` OID, it's worth noting here that `44363.47.3` is temporary and will eventually be replaced by a relative OID allocated by IANA.

bwesterb: Where should IANA allocate it? (See other allocations.) I think it might be fine to leave as is.

cjpatton: I'm suggesting structuring thee text this way:

https://ietf-plants-wg.github.io/merkle-tree-certs/draft-ietf-plants-merkle-tree-certs.html#section-5.1-7

> The attribute's value is a RELATIVE-OID containing the trust anchor ID's ASN.1 representation. For example, the distinguished name for a CA with ID `32473.1` would be represented in syntax of {{?RFC4514}} as:
> 
> ~~~
> 1.3.6.1.5.5.7.25.TBD=#0d0481fd5901
> ~~~
> 
> For initial experimentation, early implementations of this design will:
> 
> 1. Use UTF8String to represent the attribute's value rather than RELATIVE-OID. The UTF8String contains trust anchor ID's ASCII representation, e.g. `32473.1`.
> 
> 1. Use the OID 1.3.6.1.4.1.44363.47.1 instead of `id-rdna-trustAnchorID`.
> 
> For example, the distinguished name for a CA with ID `32473.1` would be represented in syntax of {{?RFC4514}} as:
> 
> ~~~
> 1.3.6.1.4.1.44363.47.1=#0c0733323437332e31
> ~~~
>

PR #275, opened 2026-07-18
Remove the subtree edge case entirely by allowing zero-length subtrees

davidben:
This includes https://github.com/ietf-plants-wg/merkle-tree-certs/pull/274

Instead of having an rare edge case where landmark intervals of size one return one subtree instead of two, just always return two. In order to do that, we have to say zero-length subtrees are legal.

That introduces silly special cases for consistency proofs, but means all our functions have uniform types, which I think is worth it.

### DIFF
--- a/draft-ietf-plants-merkle-tree-certs.md
+++ b/draft-ietf-plants-merkle-tree-certs.md
@@ -300,7 +300,7 @@ Landmark:
 : One of a sequence of tree sizes, infrequently chosen and used to calculate landmark subtrees for predistribution to relying parties.
 
 Landmark subtree:
-: One of the (up to two) subtrees determined by an interval between two landmarks. Predistributed and used as the basis for landmark-relative certificates.
+: One of the two subtrees determined by an interval between two landmarks. Predistributed and used as the basis for landmark-relative certificates.
 
 Standalone certificate:
 : A certificate containing an inclusion proof to some subtree, and several cosignatures over that subtree.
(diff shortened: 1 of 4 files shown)


draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -2339,30 +2354,30 @@ For all the test vectors, a tree `D_n` of size `n` is constructed with leaf valu
...
+For each value of `end` from 0 to 130, and each value of `start` from 0 to `end`, if `[start, end)` is a valid subtree, for each value of `index` from `start` to `end - 1`, add to the rolling hash the ASCII string `INDEX [START, END)`, then, for each hash in the inclusion proof ({{subtree-inclusion-proofs}}) for `d[index]` in the subtree `[start, end)`, a space (U+0020) followed by the hexadecimal encoding of that hash, and finally a newline (U+000A), where `INDEX` is the decimal representation of `index`, and `START` and `END` are the decimal representations of `start` and `end`, respectively.
```
davidben, line 2380: The hash is unchanged because all the extra loop iterations are are vacuous. I added them just to keep it all uniform.

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -300,7 +300,7 @@ Landmark:
 : One of a sequence of tree sizes, infrequently chosen and used to calculate landmark subtrees for predistribution to relying parties.
 
 Landmark subtree:
-: One of the (up to two) subtrees determined by an interval between two landmarks. Predistributed and used as the basis for landmark-relative certificates.
+: One of the two subtrees determined by an interval between two landmarks. Predistributed and used as the basis for landmark-relative certificates.
```
lukevalenta: Just to make it extra clear that empty/zero-length subtrees are now allowed:
```suggestion
: One of the two (possibly empty) subtrees determined by an interval between two landmarks. Predistributed and used as the basis for landmark-relative certificates.
```

lukevalenta (approved):
Oh this is a really nice simplification. LGTM.

PR #278, opened 2026-07-20
Some minor wording adjustments

ounsworth:


draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1052,23 +1052,23 @@ The fields of a TBSCertificateLogEntry are defined as follows:
...
+MTCLogEntry is an extensible structure. Future documents MAY define new values for MTCLogEntryType or MTCLogEntryExtensionType, with corresponding semantics. See {{certification-authority-cosigners}} and {{extensibility}} for additional discussion.
 
-An MTCLogEntry's size SHOULD NOT exceed 65535 (2<sup>16</sup>-1) bytes. Doing so may exceed size limits in common log-serving protocols, such as {{TLOG-TILES}}. TBSCertificateLogEntry does not include signatures and hashes public keys, so post-quantum algorithms do not contribute to this size.
+An MTCLogEntry's size SHOULD NOT exceed 65535 (2<sup>16</sup>-1) bytes. Doing so might exceed size limits in common log-serving protocols, such as {{TLOG-TILES}}, so implementer SHOULD be caution about any limits imposed by the protocls and log providers that they intend to use. TBSCertificateLogEntry does not include signatures and hashes public keys, so post-quantum algorithms do not contribute to this size.
```
davidben: caution -> cautios, protocls -> protocols.
Though "should be cautious about" reads like [SHOULD CONSIDER](https://www.rfc-editor.org/rfc/rfc6919.html#section-2). :-)

I dunno, I guess we should sort out what we're actually trying to say here. Previously it was:
* You probably don't want to exceed 65535. You probably are using tlog-tiles, and that is limited there
* PS: Don't worry about PQ stuff

And I suppose the point here is that there's an implicit "you should honor the limits or whatever other things impose on you" that's... honestly tautological but perhaps a glue thought missing between them.

But also maybe this is too complicated and we just keep it simple:

```suggestion
An MTCLogEntry's size SHOULD NOT exceed 65535 (2<sup>16</sup>-1) bytes. TBSCertificateLogEntry does not include signatures and hashes public keys, so post-quantum algorithms do not contribute to this size.
```

That:
1. Makes you compatible with tlog-tiles, but that's not normative text and the little FYI seems to be too complicated.
2. Hints that, if you're building a log protocol that's not tlog, you should maybe try to accommodate 65535

And then if hypothetically someone were building something new on top of MTCs and needed a different limit, they can sort that out then. We don't need to say "if you have other limits, you have other limits".

Thoughts?

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1052,23 +1052,23 @@ The fields of a TBSCertificateLogEntry are defined as follows:
...
+This document does not prescribe a particular method of observing the issuance log. The access protocols do not affect certificate interoperability, and different applications could have different needs. For example, a PKI that authenticates public services might publicly serve issuance logs, while a PKI that authenticates a single organization's intranet services might keep the log private to the organization which could, for example, require an authentication mechanism. Relying parties SHOULD define log serving requirements, including the allowed protocols and expected availability, as part of their policies on which CAs to support. See also {{log-availability}}.
```
davidben: Nit: I was always taught that "which" needs a comma, but 🤷 

Not sure about the second half. Talking about authenticating twice in the same sentence, but with different meaning, seems confusing. Do we need to elaborate on "private to the organization"? Whether it's auth or network-level or whatever can probably be left for people to choose.

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1181,28 +1181,30 @@ This is equivalent to the concatenation of:
 
 For example, the trust anchor ID 32473.1 would be encoded as the ASCII string `oid/1.3.6.1.4.1.32473.1`.
 
-`start` and `end` MUST define a valid subtree of the log, and `subtree_hash` MUST be the subtree's hash value in the cosigner's view of the log.
+`start` and `end` MUST define a valid subtree of the log {{sec-subtree}}, and `subtree_hash` MUST be the subtree's hash value {{sec-subtree}} in the cosigner's view of the log.
```
davidben: I think this will render as "`start` and `end` MUST define a valid subtree of the log Section 1.2.3". I.e. just stick "Section 1.2.3" directly inline in the text. I usually put parens like `({{sec-subtree}})` or write `See {{sec-subtree}}`.

Although these are both citing the same section. It's annoying that our specs don't allow more lightweight links everywhere but, failing that, perhaps:

```suggestion
`start` and `end` MUST define a valid subtree of the log , and `subtree_hash` MUST be the subtree's hash value in the cosigner's view of the log. See {{sec-subtree}}.
```

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -430,7 +430,7 @@ Not all intervals can form subtrees. Subtrees are limited to intervals that can
 
 {{subtree-test-vectors}} provides test vectors for the algorithms defined in this section.
 
-## Definition of a Subtree
+## Definition of a Subtree {#sec-subtree}
```
davidben: (This works, but we also just automatically get section references. I.e. we can already just cite {{definition-of-a-subtree}}. A bit verbose but does save the effort of coming up with two names for every section.)

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1181,28 +1181,30 @@ This is equivalent to the concatenation of:
...
+Before signing a subtree of some log, the cosigner MUST ensure that `subtree_hash` is consistent with its view of the log. Different cosigner roles will obtain this assurance differently. For example:
 
-* A cosigner may maintain a full copy of the log, e.g. if it's the log operator. The cosigner can then compute `subtree_hash` from this copy.
+* A cosigner MAY maintain a full copy of the log, e.g. if it's the log operator. The cosigner can then compute `subtree_hash` from this copy.
 
-* A cosigner may maintain the hash of the largest consistent tree observed by the log. The cosigner can then check `subtree_hash` with a subtree consistency proof ({{subtree-consistency-proofs}}).
+* A cosigner MAY maintain the hash of the largest consistent tree observed by the log. The cosigner can then check `subtree_hash` with a subtree consistency proof ({{subtree-consistency-proofs}}).
 
-In both cases, the cosigner MUST ensure that, as it updates its view of the log, the old and new views are consistent. For example, {{TLOG-WITNESS}} defines a cosigner that checks consistency proofs ({{Section 2.1.4 of !RFC9162}}) between the two views.
```
davidben, line 1201: Re losing the tlog-witness ref, straddling the "tlog is an important example, but not a requirement" line is tricky, but I've often found writing too abstract to be confusing. WDYT about adding a sentence like this to the middle bullet point.

> {{TLOG-WITNESS}} is an example of such a cosigner.

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1300,7 +1302,7 @@ A Merkle Tree Certificate is constructed from the following inputs:
...
 
-For any given TBSCertificateLogEntry, there are multiple possible certificates that may prove the entry is certified by the CA and publicly logged, varying by choice of subtree and signatures. {{certificate-format}} defines how the certificate is constructed based on those choices. {{standalone-certificates}} and {{landmark-relative-certificates}} define two profiles of Merkle Tree Certificates, standalone certificates and landmark-relative certificates, and how to select the subtree and signatures for them.
+For any given TBSCertificateLogEntry, there are multiple possible certificates that may prove the entry is certified and publicly logged by the the issuing CA indicated in its TBSCertificateLogEntry, varying by choice of subtree and signatures. {{certificate-format}} defines how the certificate is constructed based on those choices. {{standalone-certificates}} and {{landmark-relative-certificates}} define two profiles of Merkle Tree Certificates, standalone certificates and landmark-relative certificates, and how to select the subtree and signatures for them.
```
davidben: Hmm. This says something subtly different:
* That the entry is specifically publicly logged by the CA
* That you determine the issuing CA from the TBSCertificateLogEntry rather than being one of the inputs above

I think both of these aren't quite right. It *might* be that you deploy MTCs where the CA doesn't make its log available, but pushes to mirrors which do make it available. This is kinda silly, but maybe your PKI has CAs with very low serving capacity.

But the original text wasn't right either because maybe your PKI doesn't *want* to log them *publicly*.

The second is also odd. Yes, there is an `issuer` field in the log entry, but it's [not really used](https://mailarchive.ietf.org/arch/msg/plants/bwW5ljl-rdRqfRvqfZnFNCiL408/).

WDYT if we just simplified this down to:

```suggestion
By varying the choice of subtree and signatures, there can be multiple ways to prove the entry is in the log, and thus certified by the CA. {{certificate-format}} defines how a certificate is constructed based on those choices. {{standalone-certificates}} and {{landmark-relative-certificates}} define two profiles of Merkle Tree Certificates, standalone certificates and landmark-relative certificates, and how to select the subtree and signatures for them.
```

I.e. we're just referring back up to the list of inputs above.

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1382,7 +1384,7 @@ A *landmark-relative certificate* is a Merkle Tree certificate which contains no
 
 ### Landmark Tree Sizes
 
-To issue landmark-relative certificates, a CA must additionally maintain a *landmark sequence*, which is a sequence of *landmarks*.
+To issue landmark-relative certificates, a CA MUST additionally maintain a *landmark sequence*, which is a sequence of *landmarks*.
```
davidben: Elsewhere we say that the landmark sequence is optional, but this suggests it's unconditionally required. Perhaps:


```suggestion
A CA that issues landmark-relative certificates MUST additionally maintain a *landmark sequence*. A landmark sequence is a sequence of *landmarks*, defined below:
```

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -2349,7 +2351,7 @@ They are hash values of the outputs of all possible inputs for each algorithm, f
 
 For all the test vectors, a tree `D_n` of size `n` is constructed with leaf values `d[0] = 0x00, d[1] = 0x01, ...`. The hash function used is SHA-256. The hash values are encoded in hexadecimal.
 
-## Subtree Hashes
+## Subtree Hashes {#subtree-hashes}
```
davidben: (This is implicit from the section title.)

davidben (commented):
Thanks! That makes things a bit clearer! Some comments below.

draft-ietf-plants-merkle-tree-certs.md
ounsworth: Yeah, I agree this is a weird non-normative "but what about the ~children~ TLOG" sentence that mostly doesn't do anything. I'll give you another commit that goes halfway to Deleted.

draft-ietf-plants-merkle-tree-certs.md
ounsworth: Agree. Reverted.

draft-ietf-plants-merkle-tree-certs.md
ounsworth: Yup, that's better.

By the way, does the {{name-of-section}} syntax work for all kramdown, or did you have to do something in the preamble to make that work?

draft-ietf-plants-merkle-tree-certs.md
ounsworth: I'm not going to make a change here.

To me, this:

> {{TLOG-WITNESS}} is an example of such a cosigner.

doesn't add any value to the spec. To me, that reads like gratuitious marketing in a similar way that 

> ML-DSA private keys MAY be reperesented in the BOTH struct. {{openssl}} is an example of such a library

really adds nothing useful or actionable to the spec. But also, I don't really care if you want to put it back.

draft-ietf-plants-merkle-tree-certs.md
davidben: > By the way, does the {{name-of-section}} syntax work for all kramdown, or did you have to do something in the preamble to make that work?

As far as I know, it works for all of kramdown.

draft-ietf-plants-merkle-tree-certs.md
davidben: Fair enough. I was thinking less marketing here and more "people get confused when I talk in the abstract, so here's an example".

davidben: (But I don't feel strongly either way. Happy to leave it out.)

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1052,23 +1052,23 @@ The fields of a TBSCertificateLogEntry are defined as follows:
...
 
-MTCLogEntry is an extensible structure. Future documents may define new values for MTCLogEntryType or MTCLogEntryExtensionType, with corresponding semantics. See {{certification-authority-cosigners}} and {{extensibility}} for additional discussion.
+MTCLogEntry is an extensible structure. Future documents MAY define new values for MTCLogEntryType or MTCLogEntryExtensionType, with corresponding semantics. See {{certification-authority-cosigners}} and {{extensibility}} for additional discussion.
 
-An MTCLogEntry's size SHOULD NOT exceed 65535 (2<sup>16</sup>-1) bytes. Doing so may exceed size limits in common log-serving protocols, such as {{TLOG-TILES}}. TBSCertificateLogEntry does not include signatures and hashes public keys, so post-quantum algorithms do not contribute to this size.
+An MTCLogEntry's size SHOULD NOT exceed 65535 (2<sup>16</sup>-1) bytes. Implementers SHOULD CONSIDER that log-serving protocols, or the underlying transport mechanisms they use, could have lower limits. TBSCertificateLogEntry does not include signatures and hashes public keys, so post-quantum algorithms do not contribute to this size.
```
davidben: Hahaha. I'm pretty sure RFC 6919 is a joke that you're not actually meant to use. :-)

I dunno, is the issue just that it's a SHOULD NOT instead of a MUST NOT and thus things are wishy-washy? Why don't we just cut to the chase then and do:

```suggestion
An MTCLogEntry's size MUST NOT exceed 65535 (2<sup>16</sup>-1) bytes. TBSCertificateLogEntry does not include signatures and hashes public keys, so post-quantum algorithms do not contribute to this size.
```

Would that work better? Now we just unambiguously say 2^16-1 is the limit, no ambiguity about it. Would that remove the need for the awkward middle sentence for you?

davidben (commented):
One last comment but otherwise looks good. Thanks for this!

davidben:
Merged and then added you to acknowledgements in https://github.com/ietf-plants-wg/merkle-tree-certs/commit/2e87ea1cdd7ab01d7d08c12dd42714b559943834

PR #289, opened 2026-07-28
Change MTCSignature to SubtreeSignature

aarongable:
This better represents what data the signature is computed over, aligns with the body text ("`signatures` contains the chosen subtree signatures"), and reduces confusion given that this field is contained within the object which ends up acting as the signature on an MTC.


davidben:
Oh haha yes that was somewhat unfortunate of naming.

PR #299, opened 2026-08-27
Fix wording of rules related to MerkleTreeCertEntry extensions

zacikpa:
Noticed some typos/grammar issues while reading through the draft.


davidben (approved):
Ah that's *much* better. Thanks!

PR #301, opened 2026-08-28
Test vectors for large subtree inclusion proofs

### LINKED ISSUE #292: Add test vectors for maximally large trees
The accumulated tests cover all small values, but we also have to worry about overflow conditions at maximally large trees. A CA does not, but other parties may see consistency and inclusion proofs referencing large entries. The use of BIT_CEIL means that we might overflow slightly.

The good news is that overflow mostly does not exist anymore. The log/index split meant our logs are at most 2^48, not 2^64, and any integer type that people actually use can represent 2^48-1 can also represent 2^48. Nonetheless, we should:

* Have some test cases that go up to end = 2^48-1.
* Have some test cases for end = 2^64-1 anyway. While not possible in MTC, it is more generally possible and people might reuse the subtree machinery.

Footnote: we currently say log indices go up to 2^48-1. Practically they go up to 2^48-2 because otherwise you cannot express the end value in an MTCProof.

davidben:
Closes #292

~This handles subtree validity and subtree inclusion proofs. Ideally we would also pick up subtree consistency proofs and the find matching subtrees function, but I'll wrestle with that separately.~ Eh, this now has all of them.


davidben:
Updated with consistency proofs too

lukevalenta (approved):

PR #302, opened 2026-08-30
Fix X.690 reference

robstradling:
Four small fixes to the X.690 normative reference, one commit each:

1. **`ISO/IEC 8824-1:2021` → `8825-1:2021`.** 8824-1 is X.680, the ASN.1 *notation*. X.690, the *encoding rules*, is 8825-1.
2. **`encoding Rules` → `encoding rules`,** matching ITU-T's own capitalization.
3. **Add a `target` URL.** X.690 was the only non-RFC reference without one.
4. **Split the seriesinfo into name and value.** The bare scalar form emitted `name="ISO/IEC 8824-1:2021" value=""`; the empty value renders as a stray space before the comma.

Result:

    -          (DER)", ISO/IEC 8824-1:2021 , February 2021.
    +          (DER)", ISO/IEC 8825-1:2021, February 2021,
    +          <https://www.itu.int/rec/T-REC-X.690>.

### DIFF
--- a/draft-ietf-plants-merkle-tree-certs.md
+++ b/draft-ietf-plants-merkle-tree-certs.md
@@ -50,7 +50,7 @@ author:
 
 normative:
   X.690:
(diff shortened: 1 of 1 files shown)


davidben (approved):
Thanks!

PR #309, opened 2026-09-01
Fix a stale step reference in the issuance procedure

robstradling:
Step 5 referred to "the subtree cosignatures collected in steps 3 and 5", which points at itself.

Before #306 the list ran to six steps, with external subtree cosignatures collected in step 5, so the reference was correct. That PR merged the checkpoint and subtree cosignature requests into a single step 4 and renumbered the rest, but the cross-reference still names the old number.

The subtree cosignatures are now collected in step 3, where the CA signs each subtree with its own key, and step 4, where it requests cosignatures from external cosigners.


davidben (approved):

davidben:
Wow, you're really on a roll here. Thanks for all these fixes!

PR #312, opened 2026-09-01
Specify the landmark publication format precisely

robstradling:
The landmark publication format is fetched by relying parties from CAs, so
two implementations that parse it differently is an interoperability
problem rather than a local matter. The description was looser than the
surrounding text, which pins the line terminator to "a newline character
(U+000A)" and, in the test vector appendix, the separator to "a space
(U+0020)".

This makes the first line's separator exactly one U+0020, rules out
leading zeros on both lines, states the range of the values, and says what
"invalid" obliges a relying party to do.

The 2^48-1 bound is not a new constraint: landmark zero has tree size zero
and tree sizes are strictly monotonically increasing, so a landmark number
never exceeds its tree size, which Section 5.2 already caps at 2^48-1.

The second bullet is also reworded slightly to avoid "integer,
representing a tree size" reading awkwardly once "without leading zeros"
is added.

### DIFF
--- a/draft-ietf-plants-merkle-tree-certs.md
+++ b/draft-ietf-plants-merkle-tree-certs.md
@@ -1371,11 +1371,13 @@ To ensure that only active landmarks contain unexpired certificates, set `max_ac
 
 CAs SHOULD publish their active landmarks, so that relying parties can configure trusted subtrees ({{trusted-subtrees}}). The following format can be used to describe this information. The format is the following sequence of lines. Each line MUST be terminated by a newline character (U+000A):
 
-* Two space-separated non-negative decimal integers: `<last_landmark> <num_active_landmarks>`.
+* Two non-negative decimal integers, separated by a single space character (U+0020) and represented without leading zeros: `<last_landmark> <num_active_landmarks>`.
   This line MUST satisfy the following, otherwise it is invalid:
   * `num_active_landmarks <= max_active_landmarks`
   * `num_active_landmarks <= last_landmark`
-* `num_active_landmarks + 1` lines each containing a single non-negative decimal integer, representing a tree size. Numbered from zero to `num_active_landmarks`, line `i` contains the tree size for landmark `last_landmark - i`. The tree sizes MUST be strictly monotonically decreasing and less than or equal to the log's latest tree size.
+* `num_active_landmarks + 1` lines, each containing a tree size as a non-negative decimal integer without leading zeros. Numbered from zero to `num_active_landmarks`, line `i` contains the tree size for landmark `last_landmark - i`. The tree sizes MUST be strictly monotonically decreasing and less than or equal to the log's latest tree size.
+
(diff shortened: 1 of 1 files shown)


draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1371,11 +1371,13 @@ To ensure that only active landmarks contain unexpired certificates, set `max_ac
 
 CAs SHOULD publish their active landmarks, so that relying parties can configure trusted subtrees ({{trusted-subtrees}}). The following format can be used to describe this information. The format is the following sequence of lines. Each line MUST be terminated by a newline character (U+000A):
 
-* Two space-separated non-negative decimal integers: `<last_landmark> <num_active_landmarks>`.
+* Two non-negative decimal integers, separated by a single space character (U+0020) and represented without leading zeros: `<last_landmark> <num_active_landmarks>`.
```
davidben: This does make it ambiguous whether zero is "0" or "". Not sure if there's a compact way to say that. WDYT about:

```suggestion
* Two non-negative decimal integers, separated by a single space character (U+0020) and represented without extra leading zeros: `<last_landmark> <num_active_landmarks>`.
```

Another possibility is that we toss something up in "Conventions and Definitions" that says that "decimal integers" refers to [insert lots of text here].

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1371,11 +1371,13 @@ To ensure that only active landmarks contain unexpired certificates, set `max_ac
...
   This line MUST satisfy the following, otherwise it is invalid:
   * `num_active_landmarks <= max_active_landmarks`
   * `num_active_landmarks <= last_landmark`
-* `num_active_landmarks + 1` lines each containing a single non-negative decimal integer, representing a tree size. Numbered from zero to `num_active_landmarks`, line `i` contains the tree size for landmark `last_landmark - i`. The tree sizes MUST be strictly monotonically decreasing and less than or equal to the log's latest tree size.
+* `num_active_landmarks + 1` lines, each containing a tree size as a non-negative decimal integer without leading zeros. Numbered from zero to `num_active_landmarks`, line `i` contains the tree size for landmark `last_landmark - i`. The tree sizes MUST be strictly monotonically decreasing and less than or equal to the log's latest tree size.
```
davidben: Ditto re the ambiguity around zero.

davidben (commented):
Thanks!

> The landmark publication format is fetched by relying parties from CAs

Strictly speaking, it's fetched by the backend update services of relying parties, but yes. :-)

draft-ietf-plants-merkle-tree-certs.md
robstradling: Adding something in "Conventions and Definitions" makes sense to me.  There are two mentions of "decimal integers" here, as well as several instances of "decimal representations" elsewhere in the document.  I'll propose text.

robstradling: I've pushed a commit that defines the "decimal representation" of a non-negative integer, and adapted this section to refer to that.  Note that this also tightens the existing uses of "decimal representation".

davidben (approved):

PR #317, opened 2026-09-14
Update abstract for Merkle Tree certificates

aww-aww:
Revised the abstract to improve clarity and readability.


draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -217,7 +217,7 @@ informative:
 
 --- abstract
 
-This document describes Merkle Tree certificates, a new form of X.509 certificates which integrate public logging of the certificate, in the style of Certificate Transparency. The integrated design reduces logging overhead in the face of both shorter-lived certificates and large post-quantum signature algorithms, while still achieving comparable security properties to existing X.509 constructions and Certificate Transparency. Merkle Tree certificates additionally admit an optional size optimization that avoids signatures altogether, at the cost of only applying to up-to-date relying parties and older certificates.
+This document introduces Merkle Tree certificates, a new type of X.509 certificate with native public logging similar to Certificate Transparency. This design lowers logging overhead for short-lived certificates and large post-quantum signatures, while keeping security comparable to existing standards. An optional optimization can eliminate signatures entirely, though this requires up-to-date relying parties and an existing certificate to optimize.
```
davidben, line 220: > a new type of X.509 certificate with native public logging similar to Certificate Transparency

I had it cached in my mind to avoid the word "native", but I'm not sure how widespread this preference is. I found Google's documentation on this (https://developers.google.com/style/word-list#native), but that's Google's guide.

https://authors.ietf.org/en/language-and-style points to a, uh, retracted NIST document so... I've no idea what's preferred. 🤷  Probably can find another phrasing.

davidben, line 220: > an existing certificate to optimize

I don't know how to read this clause. It seems to not say much of anything. If I just got a certificate, that's an existing certificate to optimize. Or you could misread it to mean you get an existing non-MTC certificate and then optimize that, which is extremely wrong.

The old text was pointing to the fact that your certificate had to be around for a little bit first (need a landmark). That nuance seems to be lost but maybe is not important enough to have in here so we could just remove it perhaps.

PR #318, opened 2026-09-14
Revise introduction of Merkle Tree Certificates

aww-aww:
Move the portion of the intro text that explains MTCs up to the start of the introduction and add a bit more about the high benefits before getting into all of the details.


davidben:
Hmm. The text you've moved isn't the details per se but the context needed to understand the benefits. I.e. we're introducing the problem first. Now we talk about smaller logs and such but we haven't said what CT is about yet.

PR #319, opened 2026-09-14
Correct a subtree range

aww-aww:
It is not clear if the intent was to have the range be 0x500000000000, 0xd00000000000 or 0x400000000000, 0xd00000000000, but a change was needed either way.


davidben:
No, this is correct. The subtrees contain slightly more than the range. That is how we get two subtrees instead of O(log N) subtrees.

davidben:
(`[0x500000000000, 0x900000000000)` isn't even a valid subtree.)

Issue #192, opened 2026-03-01
Split out an architecture document


davidben:
Our charter says we're going to have two documents. At a glance, the existing text that probably can go in an architecture doc include:

* Much of the introduction
* Most of Overview section
* Introduction bits of the Issuance Log
* Discussion of what cosigners are and the CA's cosigner
* The guidance on how to pick relying party cosigner policies
* Text describing how this stuff fits into a broader transparency ecosystem, maybe with (non-normative) pointers to tlog and stuff. Then sections like "Publishing Logs" can probably be removed from the protocol draft, which is really focused on how certificates work and such
* Probably all of Deployment Considerations
* Other stuff?

lukevalenta:
Another thing to perhaps touch on in the architecture doc is the interaction of MTCs with offline trust anchors.

MTC CAs are similar to intermediate issuing CAs, with online keys. However, many CA operators keep their root keys offline in air-gapped environments. How does that interact with MTCs? Do we do away with offline keys altogether since the issuing keys become de-facto trust anchors anyway?

I suppose you could have an offline key sign the software update that the ships the online trust anchors to relying parties. Not sure if others have thought deeply about this yet.

(Notably, MTC's built-in batch signing is great for slow-but-online HSMs.)

sgmenda:
The sections listed above look right, I can open a PR with a skeleton document. It'll let us start moving content over.

PR #321, opened 2026-09-16
Add a skeleton architecture document

sgmenda:
This adds a skeleton for the informational architecture document. It is purely additive and does not modify the main document.

A few questions for the reviewers:

1. What should the filename and docname be?
2. Is `consensus: true` right?
3. Does the outline look right?

See #192


sgmenda:
I misclicked, this PR is not yet ready for review. Sorry for the noise.

PR #320, opened 2026-09-15
Update for draft-ietf-tls-trust-anchor-ids-05

davidben:
Updates section numbers, reworks the text to spell out the expected group patterns, and updates the demo tool to output certificate properties. https://github.com/tlswg/tls-trust-anchor-ids/pull/125 can be used to print that output while other tools are spun up.

Might be easiest reviewed as separate commits.

@bob-beck

PR #322, opened 2026-09-16
Rework the landmark file format to include an expiry

davidben:
~This sits atop #320. Just look at the topmost commit.~

Everything around landmarks is clocked on when landmarks expire. CAs need to keep track of it anyway, so it seems silly not to just define it explicitly. That also helps relying parties process the file if the CA hasn't been keeping it continuously updated second-by-second as landmarks expire.

In doing so, that lets us recenter the definition of active landmarks on landmark expiry. That solves a whole host of problems:

First, we said the CA has a max_active_landmarks parameter, but we never wrote it down. That parameter is also wrong because we set it at a per-log basis, but it needs to be summed across all CAs.

Instead, we can now define active landmarks more directly. max_active_landmarks is now merely a per-relying-party DoS limit.  RPs and CAs will need to coordinate on this value, but it's no longer a rigid parameter.

We can solve the per-log vs per-CA limit by then removing it from the discussion of the landmark file format. Instead, we just say that CAs SHOULD allocate landmarks such that the total number is within the limits of relying parties.

What happens if the limit is exceeded? We explicitly say that if you see more landmarks than your limit, you have to take the oldest active ones, not the newest ones. While I'm here, we never said how to handle the case that you can't validate the latest landmark, but you can validate an earlier one.

This lets us remove a dependency on knowing the CA's current log. The relying party just collects all the CA's active landmarks, across all logs. If that's too many, see above. That smooths over an edge case right after operational failure.

max_active_landmarks was also used in the TAI configuration, but in reality this was never necessary. All those ranges only needed to be bounded on one side. The width max_active_landmarks was really just a proxy for certificate expiration anyway.

While I'm here, define landmark subtrees for landmark zero, now that zero-length subtrees are allowed. This will never come up (landmark zero is always expired) but makes the type system more uniform.

(For completeness, I considered following up on this change with collapsing the landmark sequence from per-log to per-CA. This felt very natural but had one problem: if a log failures happens *before* the last landmark on a log, the whole thing gets wedged.)


draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1339,54 +1339,74 @@ A *landmark-relative certificate* is a Merkle Tree certificate which contains no
...
+A landmark's expiration time MUST be greater or equal to the `notAfter` time of every TBSCertificateLogEntry whose index is less than the tree size. When allocating a landmark, CAs SHOULD set the expiration time to the current time plus the CA's maximum certificate lifetime.
+
+A landmark that is not yet expired is said to be *active*. Landmark zero is never active. At any time, a log's *active landmark subtrees* are the landmark subtrees of each currently active landmark. Active landmark subtrees are predistributed to the relying party as trusted subtrees, as described in {{trusted-subtrees}}.
+
+The above conditions imply that every unexpired entry in the log is either contained some landmark subtree or was allocated sometime after the latest landmark.
```
```suggestion
The above conditions imply that every unexpired entry in the log is either contained in some landmark subtree or was allocated sometime after the latest landmark.
```

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1339,54 +1339,74 @@ A *landmark-relative certificate* is a Merkle Tree certificate which contains no
...
+
+As the issuance log grows, CAs continuously allocate new landmarks. More frequent allocation reduces landmark-relative certificate delay, while less frequent allocation reduces the size of the relying party's predistributed state. As described in {{trusted-subtrees}}, relying parties maintain some upper bound on active landmarks per CA. CAs SHOULD allocate landmarks such that the number of active landmarks, across all their logs, is within the bound for supported relying parties. {{allocating-landmarks}} gives a RECOMMENDED procedure for allocating landmarks.
+
+Mistakes in landmark sequence allocation only impact availability, not security. That is, they will not cause the relying party to accept certificates for entries the CA did not certify. However, they might cause a relying party reject some of the CA's otherwise valid landmark-relative certificates.
```
```suggestion
Mistakes in landmark sequence allocation only impact availability, not security. That is, they will not cause the relying party to accept certificates for entries the CA did not certify. However, they might cause a relying party to reject some of the CA's otherwise valid landmark-relative certificates.
```

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1523,27 +1543,31 @@ Cosigner roles are extensible without changes to certificate verification itself
...
 * The hash of the subtree
 
-Trusted subtrees for a given log are determined by its active landmark subtrees, as described in {{landmark-tree-sizes}}. Before configuring the subtrees as trusted, the relying party MUST obtain assurance that each subtree is consistent with checkpoints observed by a sufficient set of cosigners (see {{cosigners}}) to meet its cosigner requirements. It is not necessary that the cosigners have generated signatures over the specific subtrees, only that they are consistent.
+Trusted subtrees for a CA are determined by its active landmark subtrees, as described in {{landmark-tree-sizes}}. Before configuring the subtrees as trusted, the relying party MUST obtain assurance that each subtree is consistent with checkpoints observed by a sufficient set of cosigners (see {{cosigners}}) to meet its cosigner requirements. It is not necessary that the cosigners have generated signatures over the specific subtrees, only that they are consistent.
 
 This criterion can be checked given:
 
-* Some *reference checkpoint* that contains the latest landmark
+* Some *reference checkpoint* whose tree size is greater or equal to that of latest landmark
```
```suggestion
* Some *reference checkpoint* whose tree size is greater or equal to that of the latest landmark
```

lukevalenta (approved):
This is a nice simplifying change!

PR #324, opened 2026-09-18
Update MerkleTown ref to Cloudflare Radar

joshuasing:
Cloudflare merged the Merkle Town site into Cloudflare Radar - starting around early October 2025, https://ct.cloudflare.com/ redirects to Cloudflare Radar (https://radar.cloudflare.com/certificate-transparency).


davidben:
Oh nice catch!

This does raise a citation question. The citation is dated 2023-03-07 and the text says "As of June 9th, 2025". I guess 2023 must be wrong and we forgot to update it when refreshing the numbers. But I'm guessing it wasn't Cloudflare Radar yet on either date...

Alternatively, rather than philosophize over questions like these, we can just refresh the numbers and then it'll all be fine again! 😆

joshuasing:
Thanks

I was going to submit a follow-up pull request to update the statistics as well, but I wasn't sure whether to keep the existing statistics or replace them, as I think the change is quite substantial and interesting - and could be used to the point below with issuance rate increasing with short-lived certificates.

Maybe this could be done in one pass, and can update the citation date as well?

```diff
+And as of September 17th, 2026:
+
+* {{LetsEncrypt}} reported 682,000,000 active certificates for a single CA
+* {{CloudflareRadar}} reported around 9,500,000,000 unexpired certificates in CT logs, across all CAs
+* {{CloudflareRadar}} reported an issuance rate of around 591,000 certificates per hour, across all CAs
```

Sources:
- Let's encrypt active certificates: https://letsencrypt.org/stats/
- Cloudflare Radar unexpired certificates in CT logs: https://radar.cloudflare.com/explorer?dataSet=ct&groupBy=log&dt=52w&filters=expirationStatus%253DVALID (download summary response, `cat summary.json | jq '.result.main | [.[] | tonumber] | add'` - I _think_ this is correct, but would be good to double check)
- Cloudflare Radar certificate issuance rate: https://radar.cloudflare.com/explorer?dataSet=ct&filters=uniqueEntries%253Dtrue&timeCompare=1&dt=1d (unique certificates)

davidben:
Just updating them in one pass SGTM!

joshuasing:
I have created https://github.com/ietf-plants-wg/merkle-tree-certs/pull/325 to update the statistics under Size Estimate, and I have rolled this change into it. Thanks!

PR #325, opened 2026-09-18
Update size estimates section

joshuasing:
Update statistics referenced in the Size Estimates section to newer statistics collected on September 18, 2026, and recalculate size estimates.

Cloudflare merged their Merkle Town site into Cloudflare Radar around early October 2025, as such updated statistics are from Cloudflare Radar instead - I have updated the reference accordingly (previous discussion: https://github.com/ietf-plants-wg/merkle-tree-certs/pull/324).

Sources:
- Let's encrypt active certificates: https://letsencrypt.org/stats/
- Cloudflare Radar unique unexpired certificates in CT logs: https://radar.cloudflare.com/explorer?dataSet=ct&dt=52w&filters=expirationStatus%253DVALID%252CuniqueEntries%253Dtrue (download summary response, `cat summary.json | jq '.result.main.values | [.[] | tonumber] | add'`
- Cloudflare Radar certificate issuance rate: https://radar.cloudflare.com/explorer?dataSet=ct&filters=uniqueEntries%253Dtrue&timeCompare=1&dt=1d (unique certificates)


draft-ietf-plants-merkle-tree-certs.md
davidben: This date also needs to be updated.

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1412,11 +1412,11 @@ The procedure above is not specific to the CA. Any party holding a standalone ce
...
 
-* {{LetsEncrypt}} reported around 558,000,000 active certificates for a single CA
-* {{MerkleTown}} reported around 2,100,000,000 unexpired certificates in CT logs, across all CAs
-* {{MerkleTown}} reported an issuance rate of around 444,000 certificates per hour, across all CAs
+* {{LetsEncrypt}} reported around 682,000,000 active certificates for a single CA
+* {{CloudflareRadar}} reported around 9,500,000,000 unexpired certificates in CT logs, across all CAs
+* {{CloudflareRadar}} reported an issuance rate of around 591,000 certificates per hour, across all CAs
 
 The current issuance rate across the Web PKI may not necessarily be representative of the Web PKI after a transition to short-lived certificates. Assuming a certificate lifetime of 7 days, and that subscribers will update their certificates 75% of the way through their lifetime (see {{certificate-renewal}}), every certificate will be reissued every 126 hours. This gives issuance rate estimates of around 4,400,000 certificates per hour and 17,000,000 certificates per hour, for the first two values above. Note the larger estimate is across all CAs, while subtrees would only span one CA.
```
davidben: These numbers will need to be updated. Also the next two paragraphs.

davidben (commented):
> Alternatively, it could be interesting to keep the statistics from June 9th, 2025 as a reference to show how substantially certificate issuance has increased, which would also contribute to the paragraph below the statistics. If this is wanted, I am happy to update this diff accordingly.

Hmm. That seems not super necessary. These numbers are really only needed to give a rough order of magnitude for some logarithms. I think realistically anyone looking at these numbers should mentally allow for a nontrivial multiplicative increase... which is fine because we've got logarithms here.

If we wanted to make a statement about growth over time, we'd probably need to do more work to separate out cert lifetime impacts from subscriber count impacts.

draft-ietf-plants-merkle-tree-certs.md
joshuasing: Thanks, updated in https://github.com/ietf-plants-wg/merkle-tree-certs/pull/325/commits/09201dfed72ab0f0383f6e1ee0215576fd66ce2b

draft-ietf-plants-merkle-tree-certs.md
joshuasing: Oh yes, thanks. Updated in https://github.com/ietf-plants-wg/merkle-tree-certs/pull/325/commits/584b8858983c0692415a2802eb1c4edf93479b9c

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1412,17 +1412,17 @@ The procedure above is not specific to the CA. Any party holding a standalone ce
...
 The inclusion proofs in standalone and landmark-relative certificates scale logarithmically with the size of the subtree. These sizes can be estimated with the CA's issuance rate. The byte counts below assume the issuance log's hash function is SHA-256.
 
-Some organizations have published statistics which can be used to estimate this rate for the Web PKI. As of June 9th, 2025:
+Some organizations have published statistics which can be used to estimate this rate for the Web PKI. As of September 17th, 2026:
 
-* {{LetsEncrypt}} reported around 558,000,000 active certificates for a single CA
-* {{MerkleTown}} reported around 2,100,000,000 unexpired certificates in CT logs, across all CAs
-* {{MerkleTown}} reported an issuance rate of around 444,000 certificates per hour, across all CAs
+* {{LetsEncrypt}} reported around 682,000,000 active certificates for a single CA
+* {{CloudflareRadar}} reported around 9,500,000,000 unexpired certificates in CT logs, across all CAs
```
davidben: It's a bit fishy that this number increased so dramatically more than the other two. I suspect the issue is that a single certificate actually shows up in multiple logs, and in multiple forms.

Does this look right to you? I took out the "breakdown by" (seems unhelpful) and turned on unique entries.
https://radar.cloudflare.com/explorer?dataSet=ct&dt=52w&filters=expirationStatus%253DVALID%252CuniqueEntries%253Dtrue

From there, I think we want is the sum of the entire time series? Passing that summary to `jq '.result.main.values | [.[] | tonumber] | add'` gives 2,906,366,320 which... seems more plausible?

Does that look right to you?

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1412,17 +1412,17 @@ The procedure above is not specific to the CA. Any party holding a standalone ce
...
 
-Using the per-CA short lifetime estimate, if the CA mints a checkpoint every 2 seconds, standalone certificate subtrees will span around 2,500 certificates, leading to 12 hashes in the inclusion proof, or 384 bytes. Standalone certificates additionally must carry a sufficient set of signatures to meet relying party requirements.
+Using the per-CA short lifetime estimate, if the CA mints a checkpoint every 2 seconds, standalone certificate subtrees will span around 3,000 certificates, leading to 13 hashes in the inclusion proof, or 416 bytes. Standalone certificates additionally must carry a sufficient set of signatures to meet relying party requirements.
```
davidben, line 1423: Hmm. Why did 12 turn into 13? We haven't crossed a power of 2 yet. (Or am I just too sleepy right now?)

joshuasing: Hmm, I calculated this as:
```
log2 3000 = 11.5507 + 1 = 12.5507
```
rounded up to 13. Is this correct?

draft-ietf-plants-merkle-tree-certs.md
joshuasing: Yes, that seems far more plausible. It looks like I forgot to enable the unique filter, which per their help tip sounds correct:
> Whether filter out duplicate certificates and pre-certificates.

I tried finding another source to back up this number, but all similar services seem to have vastly different numbers and no way to filter to unique certificates.

I will update with this number, thank you!

joshuasing: Updated in https://github.com/ietf-plants-wg/merkle-tree-certs/pull/325/commits/60c57d9df8a1c0e144bc5deea0254f20575101bb (also bumps to Sept 18 since this data is collected as-of today now). Thanks

draft-ietf-plants-merkle-tree-certs.md
davidben: There isn't a +1 in how the other one was computed. log2(5400000) is 22.36ish. But I know I've gotten off-by-ones here in the past, so working from first principles...

An inclusion proof in a tree of size 4 has two hashes. An inclusion proof in a tree of size 6 has up to 3 hashes, though potentially less due to right edge. So it's ceil(log2(tree_size))

The tree size is not quite the same as the range size because we have to find two subtrees in a funky way. If you're in the right subtree, that subtree is strictly smaller than the range, so those inclusion proofs are bounded by ceil(log2(range_size)). Possibly smaller depending on where the split point ends up.

Of the remainder, we're essentially rounding that up to a power of 2, although I see what I actually wrote is that the excess is less than half. That's not quite a strong enough statement to get the true upper bound. Anyway, I believe the true upper bound is ceil(log2(range_size)) without a + 1.

So... I think that means it's 12, not 13? I'll take yet another pass at the invariants so that actually translates to inclusion proof sizes tidily.

davidben: > I'll take yet another pass at the invariants so that actually translates to inclusion proof sizes tidily.

Done in https://github.com/ietf-plants-wg/merkle-tree-certs/pull/326

joshuasing: Hmm, yes, I think you are correct.

I tried to write a function following [RFC9162 2.1.3.1.](https://www.rfc-editor.org/info/rfc9162/#section-2.1.3.1), which also outputs the maximum as `12` (min on right edge is `9`): https://go.dev/play/p/21SAmNKKuv1

This doesn't increase the length to `13` until the subtree size is 4097, which sounds correct?

I have fixed this in https://github.com/ietf-plants-wg/merkle-tree-certs/pull/325/commits/978da7a9870f24837bd17d34dd0a62d7d540a52c. Thank you!

joshuasing: > > I'll take yet another pass at the invariants so that actually translates to inclusion proof sizes tidily.
> 
> Done in #326

That change makes this a lot clearer. Thanks!

davidben (approved):
Thanks!

PR #326, opened 2026-09-19
Discuss subtree inclusion proof sizes more explicitly

davidben:
We talk about the selected subtrees being "efficient", but the specific notion of efficiency we care about is subtree inclusion proof size. The stated bounds actually give a slightly weaker bound on proof size than is true, because it doesn't capture that `left` is sized up to a power of 2.

Reword the discussion a bit to capture this a bit more. I didn't include all the intermediate bounds that let you derive BIT_CEIL(end - start), but hopefully it's not too hard:

The intermediate steps are that right = [mid, end) is of course bounded by not just BIT_CEIL(end - start) but end - start and really end - mid.

left = [left_start, mid) picks left_start such that left's size is exactly BIT_CEIL(mid - start). Depending on the value of mid, that is potentially tighter than BIT_CEIL(end - start).


davidben:
The motivation is because the off-by-one in https://github.com/ietf-plants-wg/merkle-tree-certs/pull/325/changes#r4053623170 confuses me all the time.

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -863,11 +865,14 @@ While one subtree can be inefficient, two subtrees are sufficient to efficiently
...
+* The disjoint union of `left`, followed by `right`, contains `[start, end)`. That is, `left.start <= start <= left.end = right.start <= end <= right.end`.
+* While `left` may contain extra elements before `start`, `right` does not contain any extra elements. That is, `end = right.end`.
+* Each subtree's size is at most `BIT_CEIL(end - start)`.
 
-* The two subtrees cover adjacent intervals. That is, `left.end = right.start`.
-* The two subtrees together contain the entire interval `[start, end)`. There are no extra entries after `end`, but there may be extra entries before `start`. That is, `left.start <= start` and `end = right.end`.
-* If the interval is not empty, the extra entries before `start` are less than half of `left`. That is, `start - left.start < left.end - start`.
+The pair of subtree hashes for `left` and `right` can support inclusion proofs for any element of `[start, end)`. The largest such inclusion proof  is no bigger than the largest inclusion proof in `MTH(D[start:end])`. Unlike `MTH(D[start:end])`, these subtree hashes can be shown consistent with the overall Merkle Tree using subtree consistency proofs.
```
davidben: Two spaces in a row...

```suggestion
The pair of subtree hashes for `left` and `right` can support inclusion proofs for any element of `[start, end)`. The largest such inclusion proof is no bigger than the largest inclusion proof in `MTH(D[start:end])`. Unlike `MTH(D[start:end])`, these subtree hashes can be shown consistent with the overall Merkle Tree using subtree consistency proofs.
```

lukevalenta (approved):

PR #330, opened 2026-09-21
Discuss overflow when representing the ID in dotted decimal

### LINKED ISSUE #327: cosigner_name/log_origin use different encoding from the rest
When writing some WIP MTC code, I noticed that CosignedMessage `cosigner_name` and `log_origin` fields have a format mismatch. While everything else TLS-part in MTC and TAI seems to use binary ID encoding, those two fields use ASCII encoding.

Doing this format conversion seems very obnoxious to implement:
* It is base conversion from 128 to 10, which is not an easy special case.
* There seems to be no explicit limit on how big the numbers can be (other than the first, which fits in 32 bits).
* This happens in extremely critical place (certificate verification).
* Performance matters, so easy brute force is out.
* It is still fallible as the output buffer can overflow
  * The encoding can be up to 1035 bytes, but there is space for only 255.
  * `>60` byte ID can overflow, `>120` byte ID will overflow.

(The log ID in issuer was recently switched to use the binary encoding.)

davidben:
This depends on https://github.com/tlswg/tls-trust-anchor-ids/pull/126. Together with that PR, closes https://github.com/ietf-plants-wg/merkle-tree-certs/issues/327.

CC @ilaril


lukevalenta (approved):

draft-ietf-plants-merkle-tree-certs.md
```diff
@@ -1145,6 +1145,8 @@ This is equivalent to the concatenation of:
 
 For example, the trust anchor ID 32473.1 would be encoded as the ASCII string `oid/1.3.6.1.4.1.32473.1`.
 
+{{Section 4 of !I-D.ietf-tls-trust-anchor-ids}} limits OID components in trust anchor ID, such that the above value fits in 2<sup>8</sup>-1 bytes. However, a general RELATIVE-OID can be arbitrarily long, with arbitrarily large components. Implementations that do not validate these bounds earlier MUST fail signature generation or verification if `cosigner_name` or `log_origin` would be too long. Implementations MAY accept larger OID components than allowed in trust anchor IDs, but MUST support OID components up to at least 2<sup>63</sup>-1 and MUST fail signature generation or verification if conversion to dotted decimal overflows any implementation limit.
```
bob-beck, line 1148: signature generation, ok.  but so now verification of a signature requires conversion to dotted decimal before I can say this is valid? 

I don't think this should be a MUST for verification

bob-beck (changes requested):

draft-ietf-plants-merkle-tree-certs.md
davidben: Oh, it already requires it. We'd have to change the signature format away from tlog's to avoid this, or make tlog's scheme special case `oid/` on some funny way. (We can tug on that string if that's what people want. That's not this PR though.)

bob-beck: > Oh, it already requires it. We'd have to change the signature format away from tlog's to avoid this, or make tlog's scheme special case `oid/` on some funny way. (We can tug on that string if that's what people want. That's not this PR though.)

ahhh ok.  well then  meh.. although, perhaps then this isn't the place to say this? (shrug) but more indifferent about it now ;)

bob-beck (approved):

